PR #2, opened 2022-03-04
VEGA-1207 Recreate warning form

GemTay:


hawx:
https://github.com/ministryofjustice/opg-sirius-lpa-frontend/pull/2/commits/5a644361a6e23fa0e2fbdfe07a20dec2dc66c8eb shows we could Cypress test the error cases, but I'm not sure whether we should

internal/server/server.go
```diff
@@ -0,0 +1,118 @@
...
+type errorVars struct {
+	SiriusURL string
+	Path      string
+	Code      int
+	Error     string
+}
+
+type unauthorizedError interface {
+	IsUnauthorized() bool
+}
```
hawx, line 88: I think this is cleaner than having `ErrUnauthorized` as every status related problem can just be a `StatusError`, and we then implement this method to force a logout

internal/sirius/create_warning_test.go
```diff
@@ -0,0 +1,153 @@
...
+						Headers: dsl.MapMatcher{"Content-Type": dsl.String("application/json")},
+					})
+			},
+			cookies: []*http.Cookie{
+				{Name: "XSRF-TOKEN", Value: "abcde"},
+				{Name: "Other", Value: "other"},
+			},
+		},
+		{
+			name: "Validation error",
```
hawx, line 65: We haven't been doing pacts for errors, should we?

docker/docker-compose.ci.yml
```diff
@@ -0,0 +1,43 @@
...
+    volumes:
+      - ../pacts:/app/pacts
+
+  puppeteer:
+    build: ./puppeteer
+    depends_on:
+      - app
+      - pact-stub
+    environment:
+      - LHCI_BUILD_CONTEXT__CURRENT_HASH=$CIRCLE_SHA1
```
gregtyler, line 32: We need to fix this in user-management too. 
```suggestion
      - LHCI_BUILD_CONTEXT__CURRENT_HASH=$GITHUB_SHA
```

internal/sirius/create_warning_test.go
gregtyler: Not sure. There's definitely value in checking the shape of the error response (to make sure we can read from `validation_errors`), but I feel a bit uncomfortable contracting specific validation scenarios, because now we're effectively checking business logic inside a consumer.

On a related note, I'd quite like to move the auth layer into the provider test suite so that we don't have to specify the cookies inside the Pact contracts. If we ever change the auth, it'll be a real PITA updating all these tests.

internal/sirius/warning_types.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"encoding/json"
+	"net/http"
+)
+
+func (c *Client) WarningTypes(ctx Context) ([]RefDataItem, error) {
+	req, err := c.newRequest(
+		ctx,
+		http.MethodGet,
+		"/api/v1/reference-data?filter=warningType",
```
gregtyler, line 12: I believe that if you use this
```suggestion
		"/api/v1/reference-data/warningType",
```
then it comes back as `[{"handle":"...","label":"..."}]` rather than `{"warningType":[{"handle":"...","label":"..."}]}` which is a little cleaner

web/template/layouts/header.gohtml
```diff
@@ -0,0 +1,28 @@
...
+        <a class="moj-header__link moj-header__link--service-name" href="#">Sirius LPA</a>
+      </div>
+      <div class="moj-header__content">
+        <nav class="moj-header__navigation" aria-label="Account navigation">
+          <ul class="moj-header__navigation-list">
+            <li class="moj-header__navigation-item">
+              <a class="moj-header__navigation-link" href="{{ sirius "/lpa" }}" aria-current="page">LPA</a>
+            </li>
+            <li class="moj-header__navigation-item">
+              <a class="moj-header__navigation-link" href="{{ sirius "/supervision" }}" aria-current="page">Supervision</a>
```
gregtyler, line 18: I don't think these should have `aria-current="page"` on them
```suggestion
              <a class="moj-header__navigation-link" href="{{ sirius "/lpa" }}">LPA</a>
            </li>
            <li class="moj-header__navigation-item">
              <a class="moj-header__navigation-link" href="{{ sirius "/supervision" }}">Supervision</a>
```

package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "opg-sirius-lpa-frontend",
```
gregtyler, line 2: This should have an associated `package-lock.json`

package.json
```diff
@@ -0,0 +1,32 @@
...
+  "version": "0.1.0",
+  "license": "MIT",
+  "scripts": {
+    "build": "webpack",
```
gregtyler, line 6: Might be worth trying to switch to esbuild some time to bring our dependencies under control, if anyone fancies trying it.

hawx: Sounds good, but I might just add a spike for it

internal/sirius/create_warning_test.go
hawx: That makes sense, I'll remove them so it is like our others

hawx: The auth stuff shouldn't cause a problem as it is all handled in pact-proxy, so we'd be able to handle all the changes there. Whether continuing to do it this way gives us any value 🤷

gregtyler: Yeah, I think handling it in pact-proxy would be better and would keep the contract focussed on data. We'd probably need to supply a header like `Sirius-Role: System Admin` to indicate who to auth as, which might also help with those tests where we want to mock different responses for different roles.

docker/docker-compose.ci.yml
```diff
@@ -0,0 +1,43 @@
...
+
+  puppeteer:
+    build: ./puppeteer
+    depends_on:
+      - app
+      - pact-stub
+    environment:
+      - LHCI_BUILD_CONTEXT__CURRENT_HASH=$GITHUB_SHA
+      - LHCI_BUILD_CONTEXT__GITHUB_REPO_SLUG=ministryofjustice/opg-sirius-lpa-frontend
+      - LHCI_GITHUB_APP_TOKEN
```
gregtyler, line 34: This still isn't working, but not a blocker since I don't think it's working on any of our repos 😁

gregtyler (approved):

PR #6, opened 2022-03-09
VEGA-1208 Detect when in an iframe

hawx:


web/assets/main.js
```diff
@@ -15,3 +15,25 @@ GOVUKFrontend.Tabs.prototype.setup = () => {};
 
 GOVUKFrontend.initAll();
 MOJFrontend.initAll();
+
+if (window.self !== window.top) {
+  document.body.className += " app-!-embedded";
```
johnnolan, line 20: document.body.classList.add('app-!-embedded'); feels more explicit but feel free to ignore, not a biggy

web/assets/main.js
```diff
@@ -15,3 +15,25 @@ GOVUKFrontend.Tabs.prototype.setup = () => {};
...
+
+  const success = document.querySelector(".moj-banner--success");
+  if (success) {
+    window.top.postMessage(
+      "form-done",
+      `${window.location.protocol}//${window.location.host}`
+    );
+  }
+
+  document.querySelectorAll("[data-app-iframe-cancel]").forEach((el) => {
```
johnnolan, line 30: How many of these elements are you expecting? Another way might be to add a document.click handler and filter events by data attribute to keep the number of handlers down in the app. This may be overkill though example here https://github.com/ministryofjustice/opg-digideps/blob/main/client/assets/javascripts/modules/googleAnalyticsEvents.js#L3

johnnolan (approved):

web/assets/main.js
hawx: Really just one, but this will nicely handle the 0 element case if we don't have a cancel button

PR #9, opened 2022-03-17
VEGA-1233 Replace webpack with esbuild and sass

hawx:


web/template/warning.gohtml
```diff
@@ -44,7 +44,7 @@
 
         <div class="govuk-button-group">
           <button class="govuk-button" data-module="govuk-button" type="submit">Save and exit</button>
-          <a data-app-iframe-cancel class="govuk-link" href="#">Cancel</a>
+          <a data-app-iframe-cancel class="govuk-link govuk-link--no-visited-state" href="#">Cancel</a>
```
hawx, line 47: just a random thing I spotted

package.json
```diff
@@ -3,7 +3,11 @@
   "version": "0.1.0",
   "license": "MIT",
   "scripts": {
-    "build": "webpack",
+    "build": "yarn build:js && yarn build:css && yarn build:images",
+    "build:js": "esbuild --bundle web/assets/main.js --minify --outfile=web/static/javascript/all.js --sourcemap",
+    "build:css": "sass --load-path . --style compressed web/assets/main.scss web/static/stylesheets/all.css",
+    "build:images": "mkdir -p web/static/assets/images && cp node_modules/govuk-frontend/govuk/assets/images/* node_modules/@ministryofjustice/frontend/moj/assets/images/* web/static/assets/images",
```
gregtyler, line 9: CSS can be handled with [a plugin](https://www.npmjs.com/package/esbuild-sass-plugin) and you can [configure external files to be copied](https://esbuild.github.io/content-types/#data-url).

I think that would make the config a bit simpler, and it wouldn't copy across images we don't use.

hawx: I don't get why we'd use a plugin over calling `sass` directly given we aren't doing anything fancy. Same with the images, how would it know what images we use when it isn't touching the template files?

If we're going down this route I'd rather leave webpack and just swap out `babel-loader` for https://github.com/privatenumber/esbuild-loader

gregtyler (approved):

PR #12, opened 2022-03-21
VEGA-1258 Add default "none" select option for warning type

hawx:


web/template/warning.gohtml
```diff
@@ -26,6 +26,7 @@
             </p>
           {{ end }}
           <select class="govuk-select {{ if .Errors.warningType }}govuk-select--error{{ end }}" id="f-warningType" name="warning-type">
```
gregtyler, line 28: Not necessary, but a cheap bit of client side validation:
```suggestion
          <select class="govuk-select {{ if .Errors.warningType }}govuk-select--error{{ end }}" id="f-warningType" name="warning-type" required>
```

gregtyler (approved):

PR #13, opened 2022-03-22
VEGA-1245 Add form to create an event

hawx:
Because we switch between using the words "event" and "note" I've tried to stick to using "note" in `internal/sirius`, because that matches the API, and "event" everywhere else.

### DIFF
--- a/internal/sirius/case.go
+++ b/internal/sirius/case.go
@@ -1,0 +1,37 @@
+package sirius
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+)
+
+type Case struct {
+	UID      string `json:"uId"`
+	CaseType string `json:"caseType"`
+}
+
+func (c *Client) Case(ctx Context, id int) (Case, error) {
+	var v Case
+
+	req, err := c.newRequest(ctx, http.MethodGet, fmt.Sprintf("/api/v1/cases/%d", id), nil)
+	if err != nil {
+		return v, err
+	}
+
+	res, err := c.http.Do(req)
+	if err != nil {
+		return v, err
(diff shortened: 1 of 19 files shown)


internal/sirius/case.go
```diff
@@ -0,0 +1,37 @@
...
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode != http.StatusOK {
+		return v, newStatusError(res)
+	}
+
+	if err := json.NewDecoder(res.Body).Decode(&v); err != nil {
+		return v, err
+	}
```
gregtyler, line 34: It'd be nice to move some of this to a shared function at some point, it's a lot of boilerplate for what could just be `return c.get(fmt.Sprintf("/api/v1/cases/%d", id))`

internal/sirius/create_note.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+type noteRequest struct {
+	Name        string    `json:"name"`
+	Description string    `json:"description"`
+	Type        string    `json:"type"`
+	File        *NoteFile `json:"file,omitempty"`
+}
+
+func (c *Client) CreateNote(ctx Context, entityID int, entityType, noteType, name, description string, file *NoteFile) error {
```
gregtyler, line 23: Might be sensible to make `entityType` an enum for clarity of what it should be

gregtyler (approved):

internal/sirius/case.go
hawx: probably true for POSTs too, I think I'll leave for next time

gregtyler (approved):

PR #35, opened 2022-04-26
Add dark mode design

gregtyler:
Adds support for a dark mode, which is applied if the `siriusTheme` cookie is set to "dark" or "accessible-dark"

I've created a new `load-classes` JavaScript file which adds the following classes:

- `js-enabled`: Always
- `app-!-html-class--embedded` and `app-!-embedded`: If the page is loaded in a frame
- `app-!-html-class--dark`: If the `siriusTheme` cookie is set as above

It blocks the rest of the page loading to ensure the styles are applied ASAP, rather than showing a flash of mis-styled content

image image

VEGA-1282 #major

### DIFF
--- a/web/assets/dark.scss
+++ b/web/assets/dark.scss
@@ -60,4 +60,18 @@ $sirius-bg-color: #29292e;
     background-color: $sirius-bg-color;
     color: govuk-colour("white");
   }
+
+  .autocomplete__menu {
+    background-color: $sirius-bg-color;
+    color: govuk-colour("white");
+  }
+
+  .autocomplete__option--odd {
+    background-color: govuk-colour("black");
+  }
+
+  .autocomplete__option--focused,
+  .autocomplete__option:hover {
+    background-color: govuk-colour("blue");
+  }
 }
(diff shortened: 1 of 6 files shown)


web/assets/dark.scss
```diff
@@ -0,0 +1,79 @@
+$sirius-bg-color: #29292e;
+
+.app-\!-html-class--dark {
+  background-color: $sirius-bg-color;
+  color: govuk-colour("white");
+  color-scheme: dark;
```
gregtyler, line 6: This makes Chrome's UI (e.g. file upload, date selector) use dark mode

web/assets/dark.scss
```diff
@@ -0,0 +1,79 @@
...
+  .govuk-heading-m,
+  .govuk-heading-s,
+  .govuk-heading-xs,
+  p,
+  .govuk-list {
+    color: govuk-colour("white") !important;
+  }
+
+  .govuk-link:link {
+    color: lighten(govuk-colour("light-blue"), 20%);
```
gregtyler, line 31: This isn't beautiful, but it ensures we're:
- Basing our colours on the Design System's core colour scheme
- Using colours which sufficiently contrast with the background

hawx (approved):

PR #41, opened 2022-05-18
VEGA 1280 auto resize text areas

GemTay:
VEGA-1280 #patch

### DIFF
--- a/web/assets/auto-resize-text-area.js
+++ b/web/assets/auto-resize-text-area.js
@@ -1,0 +1,13 @@
+const autoResizeTextArea = () => {
+    const txtarea = document.getElementsByTagName("textarea");
+    for (let i = 0; i < txtarea.length; i++) {
+        txtarea[i].setAttribute("style", "height:" + (txtarea[i].scrollHeight) + "px;overflow-y:hidden;");
+        txtarea[i].addEventListener("input", OnInput, false);
+    }
+
+    function OnInput() {
+        this.style.height = "auto";
+        this.style.height = (this.scrollHeight) + "px";
+    }
+};
+export default autoResizeTextArea;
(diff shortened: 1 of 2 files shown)


web/assets/auto-resize-text-area.js
```diff
@@ -0,0 +1,13 @@
+const autoResizeTextArea = () => {
+    const txtarea = document.getElementsByTagName("textarea");
+    for (let i = 0; i < txtarea.length; i++) {
+        txtarea[i].setAttribute("style", "height:" + (txtarea[i].scrollHeight) + "px;overflow-y:hidden;");
+        txtarea[i].addEventListener("input", OnInput, false);
+    }
+
+    function OnInput() {
+        this.style.height = "auto";
```
GemTay, line 9: auto resets the height of the text area before the changes are applied with scrollHeight

GemTay:
image

web/assets/auto-resize-text-area.js
```diff
@@ -0,0 +1,13 @@
+const autoResizeTextArea = () => {
+    const txtarea = document.querySelectorAll(".js-auto-resize-textarea");
+    for (let i = 0; i < txtarea.length; i++) {
+        txtarea[i].setAttribute("style", "height:" + (txtarea[i].scrollHeight) + "px;overflow-y:hidden;");
+        txtarea[i].addEventListener("input", OnInput, false);
+    }
+
+    function OnInput() {
+        this.style.height = "auto";
+        this.style.height = (this.scrollHeight) + "px";
```
hawx, line 10: I think we might need to add something like 3px to the height? The padding on the bottom seems quite tight and I get a scroll bar

image

web/assets/auto-resize-text-area.js
```diff
@@ -0,0 +1,13 @@
+const autoResizeTextArea = () => {
```
hawx, line 1: When you get an error on a form but have entered a lot of text the textarea doesn't grow to fit until you edit it

web/assets/auto-resize-text-area.js
GemTay: Scroll bar shouldn't be there anymore & padding should look better too now :)

web/assets/auto-resize-text-area.js
GemTay: Fixed!

gregtyler (approved):

PR #45, opened 2022-05-25
VEGA-1250 Add create relationship form

hawx:


internal/sirius/client.go
```diff
@@ -102,10 +102,15 @@ func (e StatusError) Data() interface{} {
...
 }
 
-type ValidationErrors map[string]map[string]string
+type FieldErrors map[string]map[string]string
 
 type ValidationError struct {
-	Errors ValidationErrors `json:"validation_errors"`
+	Detail string      `json:"detail"`
+	Field  FieldErrors `json:"validation_errors"`
+}
```
hawx, line 110: The endpoint for creating a person reference returns one of the simple errors with `detail` when you try to create a logically bad reference. It seemed like it would be better to add it here instead of trying to parse the response body twice. This has meant changing a few other places to support it in the error-summary template. Also:

- renaming `Errors` to `Field` made the templates read better
- I chose to add an `Any() bool` method to check if it is set; the other option would be to always parse to a `*ValidationError`, but I thought this was cleaner as we only need to check in 2 places :shrug:

gregtyler (approved):

gregtyler (approved):

PR #57, opened 2022-06-08
VEGA-1253 Add edit dates form

hawx:


internal/sirius/date_string.go
```diff
@@ -22,6 +22,10 @@ func (s *DateString) UnmarshalJSON(text []byte) error {
 	}
 	text = text[1 : len(text)-1]
 
+	// Sirius gives dates as "03\/04\/2022", which pointlessly escapes the forward
+	// slashes, we can safely remove them
+	text = bytes.ReplaceAll(text, []byte{'\\'}, []byte{})
```
gregtyler, line 27: If only we were using y-m-d 😞

internal/sirius/edit_dates_test.go
```diff
@@ -0,0 +1,94 @@
...
+					Given("I have a pending case assigned").
+					UponReceiving("A request to edit the dates").
+					WithRequest(dsl.Request{
+						Method: http.MethodPut,
+						Path:   dsl.String("/api/v1/lpas/800"),
+						Headers: dsl.MapMatcher{
+							"X-XSRF-TOKEN":        dsl.String("abcde"),
+							"Cookie":              dsl.String("XSRF-TOKEN=abcde; Other=other"),
+							"OPG-Bypass-Membrane": dsl.String("1"),
+						},
```
gregtyler, line 36: Should probably also add the body to ensure that the code is sending the right thing and that it's accepted by the provider.
```go
					Body: map[string]interface{}{
						"rejectedDate": "04/03/2022",
					},
```

hawx: umm yep, totally missed this

gregtyler (approved):

PR #58, opened 2022-06-09
Add form to link two person records

gregtyler:
Users search for the UID of the person they want to link, check the details, select a primary record (if possible) and save.

Fixes VEGA-1251 #minor

### DIFF
--- a/internal/server/link_person.go
+++ b/internal/server/link_person.go
@@ -1,0 +1,98 @@
+package server
+
+import (
+	"net/http"
+	"strconv"
+
+	"github.com/ministryofjustice/opg-go-common/template"
+	"github.com/ministryofjustice/opg-sirius-lpa-frontend/internal/sirius"
+)
+
+type LinkPersonClient interface {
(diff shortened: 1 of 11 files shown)


internal/sirius/edit_dates_test.go
```diff
@@ -40,9 +40,6 @@ func TestEditDates(t *testing.T) {
 					}).
 					WillRespondWith(dsl.Response{
 						Status: http.StatusOK,
-						Body: dsl.Like(map[string]interface{}{
-							"rejectedDate": dsl.String("04/03/2022"),
-						}),
```
gregtyler, line 45: Took this out because the body _doesn't_ include `rejectedDate` and the Pact tests are failing (not noticed previously because a webhook was broken). Since we don't actually use the rejectedDate, there's no need to reason it in the contract.

hawx (commented):
I like the UI choice

hawx (approved):

hawx (approved):

PR #65, opened 2022-06-21
Recreate unlink record form

GemTay:
Created form & updated dark mode scss to style the table correctly. Added tests

### DIFF
--- a/internal/server/unlink_person.go
+++ b/internal/server/unlink_person.go
@@ -1,0 +1,72 @@
+package server
+
+import (
+	"fmt"
+	"github.com/ministryofjustice/opg-go-common/template"
+	"github.com/ministryofjustice/opg-sirius-lpa-frontend/internal/sirius"
+	"net/http"
+	"strconv"
+)
+
+type UnlinkPersonClient interface {
+	Person(sirius.Context, int) (sirius.Person, error)
+	UnlinkPerson(sirius.Context, int, []int) error
+}
+
+type unlinkPersonData struct {
+	XSRFToken string
+	Success   bool
+	Error     sirius.ValidationError
+
+	Entity sirius.Person
+}
+
+func UnlinkPerson(client UnlinkPersonClient, tmpl template.Template) Handler {
+	return func(w http.ResponseWriter, r *http.Request) error {
+		parentID, err := strconv.Atoi(r.FormValue("id"))
+		if err != nil {
+			return err
+		}
+
+		ctx := getContext(r)
+		data := unlinkPersonData{XSRFToken: ctx.XSRFToken}
+
+		data.Entity, err = client.Person(ctx, parentID)
+		if err != nil {
+			return err
+		}
+
+		var childIds []int
+
+		if r.Method == http.MethodPost {
+			for i := range data.Entity.Children {
+				id := r.FormValue(fmt.Sprintf("child-%d", i))
+				if id != "" {
+					id, err := strconv.Atoi(id)
+					if err != nil {
+						return err
+					}
+					childIds = append(childIds, id)
+				}
+			}
+
+			if len(childIds) == 0 {
+				w.WriteHeader(http.StatusBadRequest)
+				data.Error = sirius.ValidationError{
+					Field: sirius.FieldErrors{
+						"child": {"reason": "Please select the record to be unlinked"},
+					},
+				}
+			} else {
+				err = client.UnlinkPerson(ctx, parentID, childIds)
+				if err != nil {
+					return err
+				} else {
+					data.Success = true
+				}
+			}
+		}
+
+		return tmpl(w, data)
+	}
+}
--- a/internal/server/unlink_person.go
+++ b/internal/server/unlink_person.go
@@ -17,7 +17,7 @@ type unlinkPersonData struct {
 	Success   bool
 	Error     sirius.ValidationError
 
-	Entity sirius.Person
+	Person sirius.Person
 }
 
 func UnlinkPerson(client UnlinkPersonClient, tmpl template.Template) Handler {
@@ -30,7 +30,7 @@ func UnlinkPerson(client UnlinkPersonClient, tmpl template.Template) Handler {
 		ctx := getContext(r)
 		data := unlinkPersonData{XSRFToken: ctx.XSRFToken}
 
-		data.Entity, err = client.Person(ctx, parentID)
+		data.Person, err = client.Person(ctx, parentID)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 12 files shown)


internal/server/unlink_person_test.go
```diff
@@ -0,0 +1,226 @@
...
+		Return(nil)
+
+	r, _ := http.NewRequest(http.MethodGet, "/?id=189", nil)
+	w := httptest.NewRecorder()
+
+	err := UnlinkPerson(client, template.Func)(w, r)
+	resp := w.Result()
+
+	assert.Nil(t, err)
+	assert.Equal(t, http.StatusOK, resp.StatusCode)
```
hawx, line 70: should probably add `mock.AssertExpectationsForObjects(t, client, template)` to these tests, I know we aren't very good at being consistent with it

internal/server/unlink_person_test.go
```diff
@@ -0,0 +1,226 @@
...
+
+	r, _ := http.NewRequest(http.MethodGet, "/?id=189", nil)
+	w := httptest.NewRecorder()
+
+	err := UnlinkPerson(client, template.Func)(w, r)
+
+	assert.Equal(t, expectedError, err)
+}
+
+func TestUnlinkPersonWhenValidationError(t *testing.T) {
```
hawx, line 141: this is more like `TestPostUnlinkPersonWhenChildNotSelected`, and I think there should be another for when `UnlinkPerson` does return a `ValidationError`

internal/sirius/case.go
```diff
@@ -13,6 +14,7 @@ type Case struct {
 	RegistrationDate DateString `json:"registrationDate"`
 	RejectedDate     DateString `json:"rejectedDate"`
 	WithdrawnDate    DateString `json:"withdrawnDate"`
+	Children         []Case     `json:"children"`
```
hawx, line 17: did you mean to add these fields here?

internal/sirius/unlink_person.go
```diff
@@ -0,0 +1,37 @@
...
+	if err != nil {
+		return err
+	}
+
+	resp, err := c.http.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
```
hawx, line 31: this can return a standard `{"validation_errors" ...}` body, so should probably check and unmarshal it like we do elsewhere

internal/server/unlink_person.go
```diff
@@ -0,0 +1,67 @@
...
+	Person(sirius.Context, int) (sirius.Person, error)
+	UnlinkPerson(sirius.Context, int, int) error
+}
+
+type unlinkPersonData struct {
+	XSRFToken string
+	Success   bool
+	Error     sirius.ValidationError
+
+	Entity sirius.Person
```
hawx, line 20: maybe call this `Person` since `Entity` elsewhere is used to populate a little context of what case/person you're looking at (and is hopefully a temporary solution before the full page UI is sorted out)

internal/sirius/case.go
GemTay: Whoops, no, I made this change at the start and then reverted but I think they've come back in a rebase

web/template/unlink_person.gohtml
```diff
@@ -0,0 +1,96 @@
...
+                        <td class="govuk-table__cell">{{ .Person.Firstname }} {{ .Person.Surname }}</td>
+                    </tr>
+                    <tr class="govuk-table__row">
+                        <th scope="row" class="govuk-table__header">Contact ref #</th>
+                        <td class="govuk-table__cell">{{ .Person.UID }}</td>
+                    </tr>
+                    <tr class="govuk-table__row">
+                        <th scope="row" class="govuk-table__header">Address</th>
+                        <td class="govuk-table__cell">{{ .Person.AddressLine1 }}</td>
+                    </tr>
```
gregtyler, line 29: Not a blocked, but in case it comes up: the old forms didn't ever actually show the address because they referred to an undefined variable called `uiAddress`

internal/sirius/person_test.go
```diff
@@ -46,6 +46,13 @@ func TestPerson(t *testing.T) {
 							"firstname": dsl.String("John"),
 							"surname":   dsl.String("Doe"),
 							"dob":       dsl.Term("01/01/1970", `^\d{1,2}/\d{1,2}/\d{4}$`),
+							"children": dsl.Like([]map[string]interface{}{{
+								"id":        dsl.Like(105),
+								"uId":       dsl.Term("7000-0000-0002", `7\d{3}-\d{4}-\d{4}`),
+								"firstname": dsl.String("Child"),
+								"surname":   dsl.String("One"),
+								"dob":       dsl.Term("05/05/1980", `^\d{1,2}/\d{1,2}/\d{4}$`),
+							}}),
```
gregtyler, line 55: I don't think it makes much difference but in the future probably worth using `EachLike` to be a bit more specific
```suggestion
							"children": dsl.EachLike(map[string]interface{}{
								"id":        dsl.Like(105),
								"uId":       dsl.Term("7000-0000-0002", `7\d{3}-\d{4}-\d{4}`),
								"firstname": dsl.String("Child"),
								"surname":   dsl.String("One"),
								"dob":       dsl.Term("05/05/1980", `^\d{1,2}/\d{1,2}/\d{4}$`),
							}, 1),
```

gregtyler (approved):

internal/server/unlink_person.go
```diff
@@ -0,0 +1,71 @@
...
+			return err
+		}
+
+		ctx := getContext(r)
+		data := unlinkPersonData{XSRFToken: ctx.XSRFToken}
+
+		data.Person, err = client.Person(ctx, parentID)
+		if err != nil {
+			return err
+		}
```
hawx, line 36: you should do this after handling the post, that way when the form comes back you'll have the correct state of the children shown

web/template/unlink_person.gohtml
```diff
@@ -0,0 +1,96 @@
...
+                        <th scope="row" class="govuk-table__header">Number of links</th>
+                        <td class="govuk-table__cell">{{ len .Person.Children }}</td>
+                    </tr>
+                    </tbody>
+                </table>
+
+            <form class="form" method="POST">
+                <input type="hidden" name="xsrfToken" value="{{ .XSRFToken }}"/>
+
+                <p class="govuk-heading-m">Secondary records</p>
```
hawx, line 44: I wonder if it would be nice to show a message and no buttons if there are no child records, kind of like in `delete_relationship.gohtml`

GemTay: yeah thats a good idea

hawx (approved):

PR #68, opened 2022-06-23
Add form to change a case's status

gregtyler:
Fixes VEGA-1238 #minor

### DIFF
--- a/internal/server/change_status.go
+++ b/internal/server/change_status.go
@@ -1,0 +1,78 @@
+package server
+
+import (
+	"fmt"
+	"net/http"
+	"strconv"
+
+	"github.com/ministryofjustice/opg-go-common/template"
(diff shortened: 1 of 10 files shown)


internal/server/change_status.go
```diff
@@ -0,0 +1,78 @@
...
+		}
+
+		ctx := getContext(r)
+
+		caseitem, err := client.Case(ctx, caseID)
+		if err != nil {
+			return err
+		}
+
+		AvailableStatuses, err := client.GetAvailableStatuses(ctx, caseID, caseType)
```
hawx, line 47: hate to be that guy, but can you lowercase this

hawx: also in Go they tend to not use `Get` in names so would just have `client.AvailableStatuses`

hawx (approved):

PR #76, opened 2022-07-01
VEGA-1248 Add allocate cases form

hawx:


web/template/allocate_cases.gohtml
```diff
@@ -0,0 +1,86 @@
...
+      <form class="form" method="POST">
+        <input type="hidden" name="xsrfToken" value="{{ .XSRFToken }}"/>
+
+        <div class="govuk-form-group {{ if .Error.Field.assignTo }}govuk-form-group--error{{ end }}">
+          <fieldset class="govuk-fieldset">
+            <legend class="govuk-fieldset__legend">
+              <h1 class="govuk-fieldset__heading">
+                Assign to
+              </h1>
+            </legend>
```
gregtyler, line 36: There's already a `h1` on this page, so this should just be a label
```suggestion
            <legend class="govuk-fieldset__legend govuk-fieldset__legend--m">
              Assign to
            </legend>
```

internal/sirius/allocate_cases_test.go
```diff
@@ -0,0 +1,100 @@
...
+		expectedError func(int) error
+		file          *NoteFile
+	}{
+		{
+			name: "OK",
+			setup: func() {
+				pact.
+					AddInteraction().
+					Given("I have a pending case assigned").
+					UponReceiving("A request to change the assignee of the case").
```
gregtyler, line 31: Probably worth doing a test with multiple cases being assigned at once, since that's an important (and weird) bit of the contract).

internal/server/allocate_cases.go
```diff
@@ -0,0 +1,134 @@
...
+		for _, id := range r.Form["id"] {
+			caseID, err := strconv.Atoi(id)
+			if err != nil {
+				return err
+			}
+			caseIDs = append(caseIDs, caseID)
+		}
+
+		if len(caseIDs) == 0 {
+			return errors.New("id missing")
```
gregtyler, line 49: Since this is shown to users, it would be good to make it something clearer like "no cases selected".

internal/server/allocate_cases.go
```diff
@@ -0,0 +1,134 @@
...
+		})
+
+		var casesMu sync.Mutex
+		var allocations []sirius.CaseAllocation
+
+		for _, caseID := range caseIDs {
+			caseID := caseID
+
+			group.Go(func() error {
+				caseitem, err := client.Case(ctx.With(groupCtx), caseID)
```
gregtyler, line 74: It seems a waste having to do lookups for every case just to get the case type. Maybe we could (in the future) refactor the allocation API endpoint to not require them and make this unnecessary?

hawx: The only reason I thought this was OK is that we're getting the cases to show the UID on the page of the one's selected. Although one thing I could do is not requery them for the POST call, and pass all the info in from the form

gregtyler (approved):

PR #87, opened 2022-07-27
Use a template for text inputs

hawx:


web/template/add_complaint.gohtml
```diff
@@ -49,11 +49,7 @@
           </fieldset>
         </div>
 
-        <div class="govuk-form-group {{ if .Error.Field.summary }}govuk-form-group--error{{ end }}">
-          <label class="govuk-label" for="f-summary">Title</label>
-          {{ template "errors" .Error.Field.summary }}
-          <input class="govuk-input {{ if .Error.Field.summary }}govuk-input--error{{ end }}" id="f-summary" name="summary" value="{{ .Complaint.Summary }}" />
-        </div>
+        {{ template "input" (field "summary" "Title" .Complaint.Summary .Error.Field.summary) }}
```
hawx, line 52: or could pass `.` instead of `.Error.Field.summary` and build it in the template?

hawx: also maybe the name might be better as `input` so it is obvious it is to match the template

gregtyler: I'd rather pass `.Error` than `.` so we're passing in a more defined interface and only what we need

gregtyler (approved):

PR #94, opened 2022-08-01
VEGA-1254 Mi reporting

hawx:


web/template/mi_reporting.gohtml
```diff
@@ -0,0 +1,86 @@
+{{ template "page" . }}
+
+{{ define "title" }}Mi Reporting{{ end }}
```
gregtyler, line 3: "MI" should probably be capitalised here and elsewhere in the file
```suggestion
{{ define "title" }}MI Reporting{{ end }}
```

gregtyler (approved):

PR #110, opened 2022-08-15
VEGA-1393 create payments widget

GemTay:
Created the payments widget to display all payments made on a case
Added unit & pact tests
#minor


internal/server/get_payments.go
```diff
@@ -0,0 +1,59 @@
...
+
+		total := 0.00
+		for _, p := range payments {
+			amount, err := strconv.ParseFloat(string(p.Amount), 64)
+			if err != nil {
+				return err
+			}
+			total = total + amount
+		}
+		data.TotalPaid = fmt.Sprintf("%.2f", total)
```
GemTay, line 55: has to set type `TotalPaid` to a string as this is the easier way to display the amount as 2dp. As a float, it was displaying 23.50 as 23.5

internal/sirius/payment.go
```diff
@@ -0,0 +1,36 @@
...
+	ID          int           `json:"id,omitempty"`
+	CaseID      int           `json:"case_id,omitempty"`
+	Source      PaymentSource `json:"source"`
+	Amount      FeeString     `json:"amount"`
+	PaymentDate DateString    `json:"paymentdate"`
+	Type        TypeOfPayment `json:"type"`
+	CreatedDate DateString    `json:"createddate"`
+	Locked      bool          `json:"locked,omitempty"`
+	CreatedByID int           `json:"createdby_id"`
+}
```
gregtyler, line 15: I don't think these are quite right: they should all be camelcase (e.g. `paymentDate`) and there are some I wouldn't expect the API to return (e.g. CaseID)

Also it's better not to include fields you don't use (e.g. CreatedByID) because it gives the impression they're necessary and makes them harder to remove from the API in the future.

internal/sirius/fee_string.go
```diff
@@ -0,0 +1,46 @@
+package sirius
+
+import (
+	"bytes"
+	"fmt"
+	"strconv"
+)
+
+type FeeString string
```
gregtyler, line 9: Is there a benefit to converting the amount to a string here? I feel like it would be easier to just treat it as an `int` until rendering it in the template

internal/sirius/payment_types.go
```diff
@@ -0,0 +1,30 @@
+package sirius
+
+import "errors"
+
+type PaymentType string
```
gregtyler, line 5: I don't think this is used any more?

internal/sirius/payment.go
```diff
@@ -0,0 +1,36 @@
+package sirius
+
+import "fmt"
+
+type Payment struct {
+	ID          int           `json:"id,omitempty"`
+	CaseID      int           `json:"case_id,omitempty"`
+	Source      PaymentSource `json:"source"`
```
gregtyler, line 8: I'm changing this to just be the code string (e.g. `PHONE`, `MAKE`), since the `Name` part is an implementation detail that we shouldn't be exposing.

internal/sirius/payment.go
GemTay: Yeah you're right, i'll just include those fields that I need right now and update to camel case :)

internal/sirius/fee_string.go
GemTay: Good point, its just made things more awkward. `int` it is!

internal/sirius/payment_types.go
GemTay: Whoops, i've accidentally have committed this, I created it for adding a payment, although now i've thought about it, I don't think I need it, I can just set the html option values instead eg `<option value="PHONE"`

internal/server/server.go
```diff
@@ -83,6 +84,7 @@ func New(logger Logger, client Client, templates template.Templates, prefix, sir
 	mux.Handle("/allocate-cases", wrap(AllocateCases(client, templates.Get("allocate_cases.gohtml"))))
 	mux.Handle("/assign-task", wrap(AssignTask(client, templates.Get("assign_task.gohtml"))))
 	mux.Handle("/mi-reporting", wrap(MiReporting(client, templates.Get("mi_reporting.gohtml"))))
+	mux.Handle("/get-payments", wrap(GetPayments(client, templates.Get("payments.gohtml"))))
```
gregtyler, line 87: I think `/payments` would be a better URL, but we probably need to revisit _all_ of these at some point (e.g. I'd eventually like it to be `/case/%d/payments`).

internal/sirius/date_string.go
```diff
@@ -5,34 +5,41 @@ import (
...
 )
 
 // DateString is a date in the format "YYYY-MM-DD" that will unmarshal from and
 // marshal to the Sirius format of "DD/MM/YYYY"
 type DateString string
 
+var formats = []string{
+	"2006-01-02T15:04:05-07:00", // format from payments table
+	"02/01/2006",
+}
```
gregtyler, line 18: I don't think the alternative format is needed any more? If it is, I'd suggest using `time.Time` for the struct field instead.

web/template/payments.gohtml
```diff
@@ -0,0 +1,58 @@
...
+            <p class="govuk-body"><strong>{{ if eq .Case.SubType "pfa" }}PFA{{ else if eq .Case.SubType "lpa" }}LPA{{ end }} {{ .Case.UID }}</strong></p>
+            <p class="govuk-body"><strong>Total paid: £{{ fee .TotalPaid }}</strong></p>
+
+            {{ range $i, $p := .Payments }}
+                <details class="govuk-details" data-module="govuk-details" open>
+                    <summary class="govuk-details__summary">
+                <span class="govuk-details__summary-text">
+                  Payment {{ plus1 $i }}
+                </span>
+                    </summary>
```
gregtyler, line 17: Not a serious issue, but the indenting is a bit confused here.

web/template/payments.gohtml
```diff
@@ -0,0 +1,58 @@
...
+{{ define "title" }}Payments{{ end }}
+
+{{ define "main" }}
+    <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+            <p class="govuk-body"><strong>{{ if eq .Case.SubType "pfa" }}PFA{{ else if eq .Case.SubType "lpa" }}LPA{{ end }} {{ .Case.UID }}</strong></p>
+            <p class="govuk-body"><strong>Total paid: £{{ fee .TotalPaid }}</strong></p>
+
+            {{ range $i, $p := .Payments }}
+                <details class="govuk-details" data-module="govuk-details" open>
```
gregtyler, line 12: I think these were intended to be closed by default?
```suggestion
                <details class="govuk-details" data-module="govuk-details">
```

internal/server/get_payments.go
```diff
@@ -0,0 +1,54 @@
...
+
+type GetPaymentsClient interface {
+	Payments(ctx sirius.Context, id int) ([]sirius.Payment, error)
+	Case(sirius.Context, int) (sirius.Case, error)
+}
+
+type getPaymentsData struct {
+	XSRFToken string
+	Success   bool
+	Error     sirius.ValidationError
```
gregtyler, line 18: I don't think these two are needed?
```suggestion
```

internal/server/get_payments_test.go
```diff
@@ -0,0 +1,150 @@
...
+			w := httptest.NewRecorder()
+
+			err := GetPayments(nil, nil)(w, r)
+
+			assert.NotNil(t, err)
+		})
+	}
+}
+
+func TestGetPaymentsWhenFailure(t *testing.T) {
```
gregtyler, line 91: Should also have a really similar test for when `Case()` succeeds but `Payments()` then fails

internal/sirius/date_string.go
GemTay: Yeah you're right, i'll revert the whole file back as my changes aren't needed anymore

web/template/payments.gohtml
GemTay: My bad, I think its because I copied the details example from GDS

internal/server/get_payments_test.go
GemTay: good spot!

gregtyler (approved):
Super 🎉

PR #118, opened 2022-08-25
VEGA-1393 Add payment display part 2

GemTay:
Created the "Add a payment" page and added tests


web/template/add_payment.gohtml
```diff
@@ -0,0 +1,62 @@
...
+                        <option value="ONLINE" {{ if eq .Source "ONLINE" }}selected{{ end }}>Online - Notify</option>
+                        <option value="MAKE" {{ if eq .Source "MAKE" }}selected{{ end }}>Online - Make</option>
+                    </select>
+                </div>
+
+                <div class="govuk-form-group govuk-!-display-inline-block govuk-!-padding-right-4 {{ if .Error.Field.paymentDate }}govuk-form-group--error{{ end }}">
+                    <label class="govuk-label" for="f-paymentDate">Date of payment</label>
+                    {{ template "errors" .Error.Field.paymentDate }}
+                    <input class="govuk-input date-picker {{ if .Error.Field.paymentDate }}govuk-input--error{{ end }}" type="date" id="f-paymentDate" name="paymentDate" value="{{ .PaymentDate }}"/>
+                </div>
```
GemTay, line 50: I could not reuse the `input-date.gohtml` component here because I had to add the class `govuk-!-display-inline-block` to position the todays date link alongside the date picker, which would have affected the `input-date.gohtml` component everywhere else its used

web/assets/todays-date.js
```diff
@@ -0,0 +1,12 @@
+const todaysDate = () => {
+    let selectTodayLink = document.querySelector('[data-module="select-todays-date"]');
+    let datePicker = document.querySelector('.date-picker');
```
GemTay, line 3: I'm not very good with JS so i'm not exactly sure of when/where I should be using `data-module=...` or just a custom class like `date-picker` ?

internal/server/add_payment.go
```diff
@@ -0,0 +1,83 @@
...
+	Case(sirius.Context, int) (sirius.Case, error)
+}
+
+type addPaymentData struct {
+	XSRFToken string
+	Success   bool
+	Error     sirius.ValidationError
+
+	Case        sirius.Case
+	Amount      string
```
GemTay, line 22: I have tried setting the amount to both `int` (formatted as float in the UI)  and `float64` however both introduced complexity and issues around formatting the amount to and from an int (in pence) on submission, eg on validation error. Setting it to string feels like the smoothest option

gregtyler: Ahhh, interesting, that's a good solution

internal/server/add_payment.go
```diff
@@ -0,0 +1,83 @@
...
+			PaymentDate: postFormDateString(r, "paymentDate"),
+		}
+
+		data.Case, err = client.Case(ctx, caseID)
+		if err != nil {
+			return err
+		}
+
+		if r.Method == http.MethodPost {
+			m, err := regexp.Match(`^\d*\.\d{2}$`, []byte(data.Amount))
```
gregtyler, line 48: You probably want `\d+` rather than `\d*` to ensure that there's at least one digit before the decimal (otherwise `.45`) would be allowed
```suggestion
			m, err := regexp.Match(`^\d+\.\d{2}$`, []byte(data.Amount))
```

web/assets/todays-date.js
```diff
@@ -0,0 +1,14 @@
+const todaysDate = () => {
+    const selectTodayLink = document.querySelector('[data-module="select-todays-date"]');
+    const datePicker = document.querySelector('.date-picker');
```
gregtyler, line 3: This isn't necessary, but since you mentioned it, the way the GDS components work with data-module is something like:
```html
<div data-module="select-todays-date">
    <input type="date" />
    <a>Use today's date</a>
</div>
```

So that you can then do:
```suggestion
    const modules = document.querySelectorAll('[data-module="select-todays-date"]');
    Array.from(modules).forEach(module => {
        const selectTodayLink = module.querySelector('a');
        const datePicker = module.querySelector('input[type="date"]');
        ...
    });
```
That way we can easily support multiple instances of the date selector on the same page without worrying about them crossing over with each other.

web/assets/todays-date.js
```diff
@@ -0,0 +1,14 @@
...
+
+    if (selectTodayLink !== null && datePicker !== null) {
+        selectTodayLink.addEventListener(
+            "click",
+            function () {
+                datePicker.value = new Date().toJSON().slice(0, 10)
```
gregtyler, line 9: Should prevent the default action of the click, so it doesn't trigger normal link-like behaviour:
```suggestion
            function (e) {
                e.preventDefault();
                datePicker.value = new Date().toJSON().slice(0, 10)
```

web/template/add_payment.gohtml
```diff
@@ -0,0 +1,61 @@
...
+
+                <a href="javascript:void(0)" class="govuk-link govuk-link--no-visited-state govuk-!-display-inline-block" data-module="select-todays-date">Use today's date</a>
```
gregtyler, line 51: Prefer `#` as inline JavaScript can often cause security warnings in scanning tools and occasionally in browsers. This requires the "preventDefault" I mentioned above so that it doesn't actually try to link to `/lpa/frontend/add-payment?id=1234#`
```suggestion
                <a href="#" class="govuk-link govuk-link--no-visited-state govuk-!-display-inline-block" data-module="select-todays-date">Use today's date</a>
```

internal/server/add_payment.go
GemTay: Well spotted!

web/template/add_payment.gohtml
GemTay: Ahh I didn't consider the security warnings on inline JS, that's a good point, thanks! :)

PR #122, opened 2022-09-02
VEGA 1393 edit payment widget

GemTay:
Version 1 of the edit payments widget
VEGA-1393 #minor

### DIFF
--- a/internal/server/edit_payment.go
+++ b/internal/server/edit_payment.go
@@ -39,7 +39,6 @@ func EditPayment(client EditPaymentClient, tmpl template.Template) Handler {
 		}
 
 		ctx := getContext(r)
-
 		p, err := client.PaymentByID(ctx, paymentID)
 		if err != nil {
 			return err
--- a/internal/server/edit_payment.go
+++ b/internal/server/edit_payment.go
@@ -1,0 +1,108 @@
+package server
+
+import (
+	"fmt"
+	"github.com/ministryofjustice/opg-go-common/template"
+	"github.com/ministryofjustice/opg-sirius-lpa-frontend/internal/sirius"
+	"net/http"
+	"regexp"
+	"strconv"
+)
+
+type EditPaymentClient interface {
+	EditPayment(ctx sirius.Context, payment sirius.Payment) error
+	Case(sirius.Context, int) (sirius.Case, error)
+	PaymentByID(ctx sirius.Context, id int) (sirius.Payment, error)
+}
+
+type editPaymentData struct {
+	XSRFToken string
+	Success   bool
+	Error     sirius.ValidationError
+
+	Case        sirius.Case
+	PaymentID   int
+	Amount      string
+	Source      string
+	PaymentDate sirius.DateString
+}
+
+func EditPayment(client EditPaymentClient, tmpl template.Template) Handler {
+	return func(w http.ResponseWriter, r *http.Request) error {
+		caseID, err := strconv.Atoi(r.FormValue("id"))
+		if err != nil {
+			return err
+		}
+		paymentID, err := strconv.Atoi(r.FormValue("payment"))
+		if err != nil {
+			return err
+		}
+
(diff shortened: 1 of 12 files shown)


internal/server/edit_payment.go
```diff
@@ -0,0 +1,107 @@
...
+
+		data := editPaymentData{
+			XSRFToken:   ctx.XSRFToken,
+			PaymentID:   paymentID,
+			Amount:      fmt.Sprintf("%.2f", sirius.PenceToPounds(p.Amount)),
+			Source:      p.Source,
+			PaymentDate: p.PaymentDate,
+		}
+
+		data.Case, err = client.Case(ctx, caseID)
```
GemTay, line 55: The case is only used for displaying the case uId and lpa type on the template. As the case ID is retrieved from the URL param, you _could_ change the url param to a different case id and have it display a completely unrelated case to the payment. Its not likely for anyone to do that but I don't like the fact that its possible 👎🏼

GemTay, line 55: The case is only used for displaying the case uId and lpa type on the template. As the case ID is retrieved from the URL param, you _could_ change the url param to a different case id and have it display a completely unrelated case to the payment. Its not likely for anyone to do that but I don't like the fact that its possible 👎🏼 However I cannot retrieve the case uId from the payment

internal/sirius/edit_payment_test.go
```diff
@@ -0,0 +1,105 @@
...
+			setup: func() {
+				pact.
+					AddInteraction().
+					Given("I have an lpa which has been paid for").
+					UponReceiving("A request to edit a payment").
+					WithRequest(dsl.Request{
+						Method: http.MethodPut,
+						Path:   dsl.String("/lpa-api/v1/payments/123"),
+						Body: map[string]interface{}{
+							"id":          dsl.Like(123),
```
gregtyler, line 34: Whilst it doesn't matter if the ID is sent in the request, I don't think we should include it in the contract since it isn't (shouldn't be) used by the API.

web/template/payments.gohtml
```diff
@@ -41,10 +41,21 @@
...
                                         Over the phone
                                     {{ else if eq .Source "MAKE" }}
-                                        Online - Make an LPA
+                                        Online - Make
+                                    {{ else if eq .Source "ONLINE" }}
+                                        Online - Notify
                                     {{ end }}
                                 </dd>
                             </div>
+                            {{ if or (eq .Source "PHONE") (eq .Source "ONLINE") }}
```
gregtyler, line 50: We don't use it on the Sirius side yet, but when we do we should switch this to use the `locked` property.

internal/server/edit_payment.go
```diff
@@ -0,0 +1,107 @@
...
+				return err
+			}
+
+			if !m {
+				w.WriteHeader(http.StatusBadRequest)
+				data.Error = sirius.ValidationError{
+					Field: sirius.FieldErrors{
+						"amount": {"reason": "Please enter the amount to 2 decimal places"},
+					},
+				}
```
gregtyler, line 76: Might be worth moving some of this logic into a "is amount valid" helper function so that we're not duplicating code between here and `addPayment`

internal/server/edit_payment.go
gregtyler: Good point. Can you write up a ticket in the backlog to return the case data from the Payment API endpoint and then check it matches here?

internal/server/edit_payment.go
GemTay: good point!

internal/server/edit_payment.go
GemTay: https://opgtransform.atlassian.net/jira/software/projects/VEGA/boards/170/backlog?selectedIssue=VEGA-1453

gregtyler (approved):

PR #128, opened 2022-09-07
Vega 1467 fee widget tweaks

GemTay:
Content and styling tweaks
Previously missed the fact that if you enter amount in an incorrect format and enter nothing for either source or payment date, a validation error message is shown for the amount however the source and payment date do not show any validation error messages because they were only being validated when we send the request to the API. Added in logic to correct this
VEGA-1467 #patch

### DIFF
--- a/internal/server/add_payment.go
+++ b/internal/server/add_payment.go
@@ -51,6 +51,16 @@ func AddPayment(client AddPaymentClient, tmpl template.Template) Handler {
 						"amount": {"reason": "Please enter the amount to 2 decimal places"},
 					},
 				}
+				if data.Source == "" {
+					data.Error.Field["source"] = map[string]string{
+						"reason": "Value is required and can't be empty",
+					}
+				}
+				if data.PaymentDate == "" {
+					data.Error.Field["paymentDate"] = map[string]string{
+						"reason": "Value is required and can't be empty",
+					}
+				}
 				return tmpl(w, data)
 			}
 
--- a/web/template/add_payment.gohtml
+++ b/web/template/add_payment.gohtml
@@ -18,11 +18,9 @@
             <form class="form" method="POST">
                 <input type="hidden" name="xsrfToken" value="{{ .XSRFToken }}"/>
                 <div class="govuk-form-group">
-                    <h3 class="govuk-label-wrapper">
-                        <label class="govuk-label" for="f-amount">
-                            Enter amount paid
-                        </label>
-                    </h3>
+                    <label class="govuk-label" for="f-amount">
+                        Enter amount paid
+                    </label>
                     {{ template "errors" .Error.Field.amount }}
                     <div class="govuk-input__wrapper">
                         <div class="govuk-input__prefix" aria-hidden="true">£</div>
--- a/web/template/add_payment.gohtml
+++ b/web/template/add_payment.gohtml
@@ -10,17 +10,17 @@
             {{ template "error-summary" .Error }}
 
             {{ if .Success }}
-                {{ template "success-banner" ("Payment added!") }}
+                {{ template "success-banner" ("Payment added") }}
             {{ end }}
 
-            <h1 class="govuk-heading-l app-!-embedded-hide">Add a payment</h1>
+            <h1 class="govuk-heading-l">Add a payment</h1>
 
             <form class="form" method="POST">
                 <input type="hidden" name="xsrfToken" value="{{ .XSRFToken }}"/>
                 <div class="govuk-form-group">
                     <h3 class="govuk-label-wrapper">
                         <label class="govuk-label" for="f-amount">
-                            Enter amount
+                            Enter amount paid
                         </label>
                     </h3>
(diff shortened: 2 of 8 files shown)


web/template/edit_payment.gohtml
```diff
@@ -50,9 +50,9 @@
 
                 <a href="#" class="govuk-link govuk-link--no-visited-state govuk-!-display-inline-block" data-module="select-todays-date">Use today's date</a>
 
-                <div class="govuk-button-group">
+                <div class="govuk-button-group govuk-!-padding-top-6">
                     <button class="govuk-button" data-module="govuk-button" type="submit">Save</button>
-                    <a data-app-iframe-cancel class="govuk-link govuk-link--no-visited-state" href="#">Cancel</a>
+                    <a class="govuk-link govuk-link--no-visited-state" href="{{ prefix (printf "/payments?id=%d" .Case.ID) }}">Cancel</a>
```
GemTay, line 55: Cancel can now be used to return back to the all payments page, (discussed and confirmed with Katie & Claire) otherwise users would have to go back out to actions and select the 'fees' button to get back to that screen

web/template/payments.gohtml
```diff
@@ -5,64 +5,72 @@
...
-                                </dt>
-                                <dd class="govuk-summary-list__value">
+            {{ if not .Payments }}
+            <h2 class="govuk-body govuk-!-padding-top-7 govuk-!-padding-bottom-7"><strong>There is currently no fee data available to display.</strong></h2>
+            {{ else }}
+            <table class="govuk-table my-table table__no-border">
+                <tbody class="govuk-table__body">
+                <tr class="govuk-table__row">
+                    <th scope="row" class="govuk-table__header govuk-!-font-size-24">Total paid:</th>
+                    <td class="govuk-table__cell govuk-!-font-size-24"><strong>£{{ fee .TotalPaid }}</strong></td>
```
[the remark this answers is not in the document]
GemTay: ...which only makes a minor difference
image

web/template/add_payment.gohtml
```diff
@@ -10,17 +10,17 @@
...
+                {{ template "success-banner" ("Payment added") }}
             {{ end }}
 
-            <h1 class="govuk-heading-l app-!-embedded-hide">Add a payment</h1>
+            <h1 class="govuk-heading-l">Add a payment</h1>
 
             <form class="form" method="POST">
                 <input type="hidden" name="xsrfToken" value="{{ .XSRFToken }}"/>
                 <div class="govuk-form-group">
                     <h3 class="govuk-label-wrapper">
```
gregtyler, line 21: This `<h3>` tag isn't necessary and might confuse screenreaders (same on edit_payment.gohtml)

web/template/payments.gohtml
```diff
@@ -5,64 +5,72 @@
...
+                </tr>
+                </tbody>
+            </table>
+
+            <h3 class="govuk-heading-m govuk-!-padding-top-3 govuk-!-padding-bottom-2"><strong>Fee details</strong></h3>
```
gregtyler, line 22: This should be a `<h2>` and doesn't need the `<strong>` tag:
```suggestion
            <h2 class="govuk-heading-m govuk-!-padding-top-3 govuk-!-padding-bottom-2">Fee details</h2>
```

internal/server/add_payment.go
```diff
@@ -51,6 +51,16 @@ func AddPayment(client AddPaymentClient, tmpl template.Template) Handler {
...
+				if data.Source == "" {
+					data.Error.Field["source"] = map[string]string{
+						"reason": "Value is required and can't be empty",
+					}
+				}
+				if data.PaymentDate == "" {
+					data.Error.Field["paymentDate"] = map[string]string{
+						"reason": "Value is required and can't be empty",
+					}
+				}
```
gregtyler, line 63: I'm not a big fan of this because it duplicates validation and doesn't cover every option (e.g. if you enter a bad amount and an invalid source, it will only tell you about the amount) but I can't think of a better option right now.

web/template/add_payment.gohtml
```diff
@@ -10,17 +10,17 @@
...
                         <label class="govuk-label" for="f-amount">
-                            Enter amount
+                            Enter amount paid
```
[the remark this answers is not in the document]
GemTay: Yes, oops!

web/template/payments.gohtml
GemTay: This is what happens when you have a copy and paste everything, sorry!

internal/server/add_payment.go
GemTay: i'm with you, I don't like this at all but it was all I could think of :(

gregtyler: Eurgh, the wonders of Laminas doing its filter annotations before its validation ones 🙄 Oh well, guess that means it's less of a problem 😂

gregtyler (dismissed):

gregtyler (approved):

PR #131, opened 2022-09-12
VEGA-1453 Retrieve case data from the payment api endpoint

hiboabd:
Case id is now returned along with the payment information when retrieving a payment by id. This is to prevent users from potentially being able to amend the url param to return payment data for a different case and display unrelated case data for the payment.

#minor

### DIFF
--- a/internal/server/edit_payment.go
+++ b/internal/server/edit_payment.go
@@ -28,7 +28,7 @@ type editPaymentData struct {
 
 func EditPayment(client EditPaymentClient, tmpl template.Template) Handler {
 	return func(w http.ResponseWriter, r *http.Request) error {
-		caseID, err := strconv.Atoi(r.FormValue("id"))
+		_, err := strconv.Atoi(r.FormValue("id"))
 		if err != nil {
 			return err
 		}
@@ -51,7 +51,7 @@ func EditPayment(client EditPaymentClient, tmpl template.Template) Handler {
 			PaymentDate: p.PaymentDate,
 		}
 
-		data.Case, err = client.Case(ctx, caseID)
+		data.Case, err = client.Case(ctx, p.Case.ID)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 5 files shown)


internal/server/edit_payment.go
```diff
@@ -28,7 +28,7 @@ type editPaymentData struct {
 
 func EditPayment(client EditPaymentClient, tmpl template.Template) Handler {
 	return func(w http.ResponseWriter, r *http.Request) error {
-		caseID, err := strconv.Atoi(r.FormValue("id"))
+		_, err := strconv.Atoi(r.FormValue("id"))
```
hiboabd, line 31: I've left this in to validate the url parameter although the parameter is no longer used

internal/sirius/payment.go
```diff
@@ -10,6 +10,7 @@ type Payment struct {
 	Source      string     `json:"source"`
 	Amount      int        `json:"amount"`
 	PaymentDate DateString `json:"paymentDate"`
+	Case        *Case      `json:"case,omitempty"`
```
gregtyler, line 13: I'm still not very confident about pointers in Go, but we use direct references elsewhere so I'd assume we should do the same here:
```suggestion
	Case        Case       `json:"case,omitempty"`
```

internal/server/edit_payment.go
GemTay: I dont think this shouldn't be needed anymore, we should only expect the payment ID url param now :)

internal/sirius/payment.go
hiboabd: Thanks i'm not that knowledgable on pointers but `omitempty` doesn't work with structs as they don't have any [empty value](https://www.sohamkamani.com/golang/omitempty/) - and the article uses pointers as a solution to the problem

gregtyler: Oooo yes, very clever 👍🏽

GemTay (approved):

PR #133, opened 2022-09-13
Abort overlapping fetch requests

gregtyler:
When using an autocomplete: if a fetch request is already in motion when a new one is issued, then abort it.

This will prevent race conditions where earlier requests land after newer ones.

Fixes VEGA-1474 #patch

### DIFF
--- a/web/assets/select.js
+++ b/web/assets/select.js
@@ -33,14 +33,22 @@ function enhanceElement(element, source) {
   }
 }
 
-function fetchUser(prefix) {
+function fetchForAutocomplete(url, mapFunction) {
+  let controller = { abort: () => {} };
+  const fetchOptions = {};
+
   return (query, callback) => {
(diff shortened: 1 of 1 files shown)


web/assets/select.js
```diff
@@ -34,8 +34,18 @@ function enhanceElement(element, source) {
...
   return (query, callback) => {
-    fetch(`${prefix}/search-users?q=${encodeURIComponent(query)}`)
+    controller.abort();
+
+    if ("AbortController" in window) {
+      controller = new AbortController();
+      fetchOptions.signal = controller.signal;
+    }
+
+    fetch(`${prefix}/search-users?q=${encodeURIComponent(query)}`, fetchOptions)
```
hiboabd, line 48: Could a helper function be extracted here - looks like the functions are the same aside from the `search-users` the url

gregtyler: Good idea, I've extracted them out now 👍🏽

hiboabd (approved):

PR #137, opened 2022-09-15
VEGA-1448 Allow payments by cheque

GemTay:
Enabled cheque payments to be editable by using the Locked field on payments.
Small refactoring tweaks to edit payment url param and test fixes
VEGA-1448 #minor

### DIFF
--- a/internal/server/edit_payment.go
+++ b/internal/server/edit_payment.go
@@ -30,7 +30,7 @@ type editPaymentData struct {
 
 func EditPayment(client EditPaymentClient, tmpl template.Template) Handler {
 	return func(w http.ResponseWriter, r *http.Request) error {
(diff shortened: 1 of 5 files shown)


internal/server/edit_payment.go
```diff
@@ -30,7 +30,7 @@ type editPaymentData struct {
 
 func EditPayment(client EditPaymentClient, tmpl template.Template) Handler {
 	return func(w http.ResponseWriter, r *http.Request) error {
-		paymentID, err := strconv.Atoi(r.FormValue("payment"))
```
GemTay, line 33: Just wanted to simplify the url param name now that we only need the payment ID

gregtyler (approved):

PR #139, opened 2022-09-21
VEGA-1476 Allow payments to be deleted

GemTay:
Added payments can now be deleted.
VEGA-1476 #minor

### DIFF
--- a/web/template/payments.gohtml
+++ b/web/template/payments.gohtml
@@ -56,6 +56,13 @@
                                     </a>
                                 </th>
                             </tr>
+                            <tr class="govuk-table__row">
+                                <th scope="row" class="govuk-table__header table-row__no-border">
+                                    <a class="govuk-link govuk-link--no-visited-state" href="{{ prefix (printf "/delete-payment?id=%d" .ID) }}">
+                                        Delete payment
+                                    </a>
+                                </th>
+                            </tr>
                             {{ end }}
                         </tbody>
                     </table>
--- a/cypress/e2e/delete-payment.cy.js
+++ b/cypress/e2e/delete-payment.cy.js
@@ -1,0 +1,13 @@
+describe("Delete a payment", () => {
+    beforeEach(() => {
+        cy.visit("/delete-payment?id=123");
+    });
+
+    it("deletes a payment on a case", () => {
+        cy.contains("Delete payment");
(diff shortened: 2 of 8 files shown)


web/assets/main.scss
```diff
@@ -26,3 +26,7 @@ $moj-page-width: 1220px;
 .table-row__no-border {
    border-bottom: none;
  }
+
+.colour-text-red {
+   color: govuk-colour("red") !important;
```
GemTay, line 31: Is there another way to override styles without using `!important` which I know should be avoided. Without it, the style was being overridden by `.govuk-link--no-visited-state:visited`

web/template/payments.gohtml
```diff
@@ -56,6 +56,13 @@
                                     </a>
                                 </th>
                             </tr>
+                            <tr class="govuk-table__row">
+                                <th scope="row" class="govuk-table__header table-row__no-border">
```
gregtyler, line 60: Probably best to add `colspan="2"` here so it spans across both columns, since it's full width.
```suggestion
                                <th scope="row" class="govuk-table__header table-row__no-border" colspan="2">
```

web/assets/main.scss
```diff
@@ -26,3 +26,7 @@ $moj-page-width: 1220px;
 .table-row__no-border {
    border-bottom: none;
  }
+
+.colour-text-red {
```
gregtyler, line 30: A couple of conventions I'd suggest using here:
- Add `app-` as a prefix to clarify who the style belongs to (compared to `govuk-` and `moj-`)
- Use `!-` as a secondary prefix to highlight that this uses `!important` and will override other rules (this is also [from the Design System](https://design-system.service.gov.uk/styles/layout/#override-how-elements-display))
```suggestion
.app-!-colour-text-red {
```

web/assets/main.scss
gregtyler: I think it's fine in this context but I've added a comment about naming it to be a bit clearer.

An alternative would be updating the selector to also target pseudoselectors:
```.colour-text-red, .colour-text-red:visited {```
But I suspect you'd quickly find you need to do all of them:
```.colour-text-red, .colour-text-red:visited, .colour-text-red:active, .colour-text-red:hover {```

gregtyler (approved):

PR #141, opened 2022-09-21
VEGA-1465 create apply fee reduction form

hiboabd:
Allows users with the Reduced Fee User role to add a fee reduction payment using the fee reduction form

#minor


internal/sirius/add_payment.go
```diff
@@ -47,3 +41,29 @@ func (c *Client) AddPayment(ctx Context, caseID int, amount int, source string,
 
 	return nil
 }
+
+func GetPostData(amount int, source string, paymentDate DateString, feeReductionType string, paymentEvidence string, appliedDate DateString) ([]byte, error) {
```
hiboabd, line 45: This is only exportable for testing purposes

web/template/payments.gohtml
```diff
@@ -68,6 +68,11 @@
             <a class="govuk-button govuk-button--secondary" href="{{ prefix (printf "/add-payment?id=%d" .Case.ID) }}">
                 Add payment
             </a>
+            {{ if .IsReducedFeesUser }}
+                <a class="govuk-button govuk-button--secondary" href="{{ prefix (printf "/apply-fee-reduction?id=%d" .Case.ID) }}">
+                    Apply fee reduction
+                </a>
+            {{ end }}
```
hiboabd, line 75: Not sure how to test this in cypress.. doesn't seem like I can control what user I am and the role(s) that user would have

internal/sirius/get_user_details_test.go
```diff
@@ -0,0 +1,83 @@
...
+					WithRequest(dsl.Request{
+						Method: http.MethodGet,
+						Path:   dsl.String("/lpa-api/v1/users/current"),
+					}).
+					WillRespondWith(dsl.Response{
+						Status: http.StatusOK,
+						Body: dsl.Like(map[string]interface{}{
+							"id":          dsl.Like(103),
+							"displayName": dsl.String("Test User"),
+							"roles":       dsl.Like([]string{"OPG User", "Case Manager", "Reduced Fees User"}),
```
hiboabd, line 41: I wanted to add a cypress test that also checks that the apply fee reduction button is not shown for users without the reduced fees user role but as i can tell i can't add another mock for this same endpoint..

internal/server/add_payment.go
```diff
@@ -18,13 +18,19 @@ type addPaymentData struct {
...
-	PaymentSources []sirius.RefDataItem
+	Case              sirius.Case
+	Amount            string
+	Source            string
+	PaymentEvidence   string
+	FeeReductionType  string
+	PaymentDate       sirius.DateString
+	AppliedDate       sirius.DateString
+	PaymentSources    []sirius.RefDataItem
+	FeeReductionTypes []sirius.RefDataItem
```
GemTay, line 29: I think it would be best to keep payments and fee reductions separate to keep each handler as simple as possible. We're also going to be adding refunds as well in the near future so if we combine everything in this handler its going to get a little messy and hard to read

internal/sirius/add_payment.go
```diff
@@ -7,16 +7,10 @@ import (
...
-		Source      string     `json:"source"`
-		PaymentDate DateString `json:"paymentDate"`
-	}{
-		Amount:      amount,
-		Source:      source,
-		PaymentDate: paymentDate,
-	})
+const FeeReductionSource = "FEE_REDUCTION"
+
+func (c *Client) AddPayment(ctx Context, caseID int, amount int, source string, paymentDate DateString, feeReductionType string, paymentEvidence string, appliedDate DateString) error {
```
GemTay, line 12: Again I think we should separate these out and have a `add_fee_reduction.go` in both `internal/sirius` and `internal/server`, as the only common field used in adding payments and fee reductions is `source` theres not much point in combining them. Therefore I think we should also create a new endpoint in sirius to add fee reductions

internal/server/add_payment.go
hiboabd: Happy to create new handlers i agree it could get messy quite fast

internal/sirius/add_payment.go
hiboabd: Happy to separate the handlers as I agree it could get messier with future work however, with the endpoints the plan is to reuse the payments endpoints with conditional logic for the different types of payments so we can easily perform calculations in the future

cypress/e2e/payments.cy.js
```diff
@@ -0,0 +1,45 @@
...
+        });
+
+        it("displays payment information if there is a payment on the case", () => {
+            cy.contains("7000-0000-0000");
+            cy.contains("Total paid");
+            cy.contains("£41.00");
+            cy.contains("Fee details");
+            cy.contains("Payments");
+            cy.get('.govuk-details__summary-text').click();
+            cy.get(".govuk-table__body > :nth-child(1) > .govuk-table__header").contains(
```
GemTay, line 22: You should be able to simplify these to `cy.contains("Amount");` `cy.contains("£41.00");` etc

hiboabd: I thought this might be better to ensure the ordering of info in the table doesn't regress in the future?

GemTay: I think we tend not to test the design or content too much because it can often be changed, whereas if we just assert that the payment data is visible we don't have to update tests when content or design changes

internal/server/add_payment_test.go
```diff
@@ -285,4 +285,4 @@ func TestPostAddPaymentAmountIncorrectFormat(t *testing.T) {
 			mock.AssertExpectationsForObjects(t, client, template)
 		})
 	}
-}
+}
```
GemTay, line 288: Sorry this is a picky one, but if you could add the blank EOF line back in it'd just keep the linters happy :) Same in `add_payment.go`

internal/sirius/get_user_details_test.go
GemTay: Arg yeah thats a pain...

internal/sirius/get_user_details_test.go
```diff
@@ -0,0 +1,110 @@
...
+				DisplayName: "Test User",
+				Roles:       []string{"OPG User", "Case Manager"},
+			},
+		},
+		{
+			name: "OK",
+			setup: func() {
+				pact.
+					AddInteraction().
+					Given("A donor exists - logged in as the reduced fees user").
```
gregtyler, line 57: I'd suggest changing this to:
```suggestion
					Given("I am a reduced fees user").
```

Then adding a matching block on the Sirius side telling it to log in as reduced.fees.user if it gets that setup state (just like the one for [Lay Team user](https://github.com/ministryofjustice/opg-sirius/blob/69cd997f4b053054f61c2706ab09c9e54ee98611/front-end/docker/pact-proxy/setup.go#L107-L110)).

I don't think you need the other Pact test above

hiboabd: Updated 👍

gregtyler (approved):

internal/server/apply_fee_reduction.go
```diff
@@ -0,0 +1,70 @@
...
+
+		data.FeeReductionTypes, err = client.RefDataByCategory(ctx, sirius.FeeReductionTypeCategory)
+		if err != nil {
+			return err
+		}
+
+		if r.Method == http.MethodPost {
+			data.Source = "FEE_REDUCTION"
+
+			err = client.ApplyFeeReduction(ctx, caseID, data.Source, data.FeeReductionType, data.PaymentEvidence, data.PaymentDate)
```
GemTay, line 57: Since data.Source will always be `"FEE_REDUCTION"` then we don't need it to be a part of the `applyFeeReductionData` struct and can just pass the string directly into the function. I could be worth setting it as a const and using that elsewhere too

hiboabd: Good point - amended

GemTay (approved):

internal/sirius/payment.go
```diff
@@ -11,12 +11,12 @@ type PaymentReference struct {
...
-	Amount      int                `json:"amount"`
-	PaymentDate DateString         `json:"paymentDate"`
-	Case        *Case              `json:"case,omitempty"`
-	Locked      bool               `json:"locked,omitempty"`
+	ID          int        `json:"id,omitempty"`
+	Source      string     `json:"source,omitempty"`
+	Amount      int        `json:"amount,omitempty"`
+	PaymentDate DateString `json:"paymentDate,omitempty"`
+	Case        *Case      `json:"case,omitempty"`
+	Locked      bool       `json:"locked,omitempty"`
```
gregtyler, line 19: Similarly, I believe your IDE should have indented all of these to match the references line, so you might want to check Go formatting is enabled

internal/sirius/add_payment.go
```diff
@@ -46,4 +46,4 @@ func (c *Client) AddPayment(ctx Context, caseID int, amount int, source string,
 	}
 
 	return nil
-}
+}
```
gregtyler, line 49: Your IDE should probably be adding newlines to the end of files, and this one shouldn't have been removed.

internal/server/get_payments.go
```diff
@@ -12,16 +12,19 @@ type GetPaymentsClient interface {
...
-	Case           sirius.Case
-	Payments       []sirius.Payment
-	PaymentSources []sirius.RefDataItem
-	ReferenceTypes []sirius.RefDataItem
-	TotalPaid      int
+	Case              sirius.Case
+	Payments          []sirius.Payment
+	PaymentSources    []sirius.RefDataItem
+	ReferenceTypes 	  []sirius.RefDataItem
+	User              sirius.User
```
gregtyler, line 25: I don't think you need the user in the template data?
```suggestion
	User              sirius.User
```

gregtyler (approved):

PR #143, opened 2022-09-26
WIP: support for refunds

gregtyler:
For VEGA-1413 #minor

### DIFF
--- a/internal/server/get_payments.go
+++ b/internal/server/get_payments.go
@@ -1,10 +1,11 @@
 package server
 
 import (
-	"github.com/ministryofjustice/opg-go-common/template"
-	"github.com/ministryofjustice/opg-sirius-lpa-frontend/internal/sirius"
 	"net/http"
 	"strconv"
+
+	"github.com/ministryofjustice/opg-go-common/template"
+	"github.com/ministryofjustice/opg-sirius-lpa-frontend/internal/sirius"
 )
 
 type GetPaymentsClient interface {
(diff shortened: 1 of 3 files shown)


hiboabd (dismissed):

hiboabd (dismissed):

web/template/layout/refund.gohtml
```diff
@@ -20,8 +20,14 @@
                     <tr class="govuk-table__row">
                         <th scope="row" class="govuk-table__header">Method of refund:</th>
                         <td class="govuk-table__cell">
-                            {{ range $.PaymentSources }}
-                                {{ if eq .Handle $p.Source }}{{ .Label }}{{ end }}
+                            {{ if eq "MAKE" $p.Source }}
+                                GOV.UK Pay (online - Make an LPA)
+                            {{ else if eq "ONLINE" $p.Source }}
+                                GOV.UK Pay (card payment)
+                            {{ else }}
```
gregtyler, line 27: I think it would be better to do this in the PaymentSource RefData in the API, then we have a single source of language and don't need to do translations like this in every frontend. But that might be easier as a follow-up ticket since the current translations are all "Paid X" so we either need to add a second translation property (like `isSelectable`) or rework the translations to be more reusable

GemTay: Good shout, i've created a ticket here: https://opgtransform.atlassian.net/jira/software/projects/VEGA/boards/170/backlog?selectedIssue=VEGA-1712

GemTay (approved):

PR #151, opened 2022-09-30
Vega 1463 edit fee reduction

GemTay:
VEGA-1463 #minor


internal/sirius/edit_payment_test.go
```diff
@@ -66,3 +66,70 @@ func TestEditPayment(t *testing.T) {
...
+		name          string
+		setup         func()
+		expectedError func(int) error
+	}{
+		{
+			name: "OK",
+			setup: func() {
+				pact.
+					AddInteraction().
+					Given("I have an lpa which has a payment and fee reduction").
```
hiboabd, line 86: This should work if you change the step to `I have an lpa which has a fee reduction` and the path should be `/lpa-api/v1/payments/124` and the cypress test needs to be `124`

internal/sirius/payment_test.go
```diff
@@ -273,3 +273,74 @@ func TestPaymentByID(t *testing.T) {
...
+						Headers: dsl.MapMatcher{"Content-Type": dsl.String("application/json")},
+					})
+			},
+			expectedResponse: Payment{
+				ID:               124,
+				PaymentEvidence:  "Some fee reduction evidence",
+				FeeReductionType: "REMISSION",
+				Source:           "FEE_REDUCTION",
+				PaymentDate:      DateString("2022-01-23"),
+				Case:             &Case{ID: 801},
```
hiboabd, line 322: The `I have an lpa which has a fee reduction` step returns a case with an id of `802` so might be good to change for consistency

hiboabd (approved):

PR #158, opened 2022-10-05
VEGA-1470 Confirm success of payment and fee reduction actions

GemTay:


web/template/add_payment.gohtml
```diff
@@ -10,7 +10,16 @@
             {{ template "error-summary" .Error }}
 
             {{ if .Success }}
-                {{ template "success-banner" ("Payment added") }}
```
GemTay, line 13: I wanted to continue using the `"success-banner"` template, however I would have had to adapt it to allow me to pass in multiple parameters (see approved answer https://stackoverflow.com/questions/18276173/calling-a-template-with-several-pipeline-parameters/18276968#18276968) so that I could pass it the "Please clear the task..." message and add logic to the template to display that if passed in - But this felt like overkill for only 4 changes

gregtyler (approved):

PR #163, opened 2022-10-20
VEGA-1441 Recreate investigation form

GemTay:
VEGA-1441 #minor

### DIFF
--- a/internal/server/create_investigation.go
+++ b/internal/server/create_investigation.go
@@ -1,0 +1,70 @@
+package server
+
+import (
+	"github.com/ministryofjustice/opg-go-common/template"
+	"github.com/ministryofjustice/opg-sirius-lpa-frontend/internal/sirius"
+	"net/http"
+	"strconv"
+)
+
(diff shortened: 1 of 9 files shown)


internal/server/create_investigation.go
```diff
@@ -0,0 +1,70 @@
...
+	return func(w http.ResponseWriter, r *http.Request) error {
+		caseID, err := strconv.Atoi(r.FormValue("id"))
+		if err != nil {
+			return err
+		}
+
+		caseType, err := sirius.ParseCaseType(r.FormValue("case"))
+		if err != nil {
+			return err
+		}
```
hiboabd, line 33: `caseType` is only used for post requests so this can be moved to the if branch on line 47

GemTay: True, however it ensures the URL contains all the necessary params for when the POST is submitted. Otherwise it could be confusing to users if they entered the form data and an error appeared on submit

hiboabd (approved):

PR #168, opened 2022-10-26
Improve success messages

gregtyler:
Two changes here:
- Rather than reloading the page whenever a success banner is found, look for the more explicit `data-app-reload` attribute containing the word "page". The existing forms have been updated to match, but in the future we might want to add flexibility so you can do e.g. `data-app-reload="timeline tasks"`
- In payments forms, instead of showing success messages in-place they now get saved as flash messages and the user is returned to `/payments?id=#` where they're shown

Fixes VEGA-1513 #minor

### DIFF
--- a/internal/server/flash.go
+++ b/internal/server/flash.go
@@ -1,0 +1,58 @@
+package server
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"net/http"
+	"time"
+)
+
+const flashCookieName string = "flash-lpa-frontend"
+
+type FlashNotification struct {
+	Title       string `json:"name"`
+	Description string `json:"description"`
+}
+
+func SetFlash(w http.ResponseWriter, notification FlashNotification) {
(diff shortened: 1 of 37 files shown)


internal/server/flash.go
```diff
@@ -0,0 +1,58 @@
...
+	str, err := json.Marshal(&notification)
+	if err != nil {
+		return
+	}
+
+	c := &http.Cookie{
+		Name:     flashCookieName,
+		Value:    base64.URLEncoding.EncodeToString(str),
+		HttpOnly: true,
+	}
```
gregtyler, line 27: Snyk is right to report that this should have `Secure: true`. However, that causes all the Cypress tests to fail because it ignores secure cookies in insecure (HTTP) contexts. Since the secure-ness of warning messages isn't particularly important, I've removed the secure flag until we can find a workaround for Cypress.

GemTay (approved):
Awesome!

PR #171, opened 2022-11-01
VEGA-1531 recreate put investigation on hold form

GemTay:
VEGA-1531 #minor


GemTay:
image
I added a summary table with the info of the investigation to be put on hold, as the original design only contained the radio buttons. This and the "Take investigation off hold" form should be redesigned properly at some point, but this is just a temporary improvement for now. I will check this is okay with Katie when she's back

hawx (approved):

PR #175, opened 2022-11-02
VEGA-1532 recreate take investigation off hold form

GemTay:
Recreated the take investigation off hold form
VEGA-1532 #minor

### DIFF
--- a/web/template/investigation_off_hold.gohtml
+++ b/web/template/investigation_off_hold.gohtml
@@ -1,0 +1,42 @@
+{{ template "page" . }}
+
+{{ define "title" }}{{ if .Error.Any }}Error: {{ end }}Investigation off hold{{ end }}
+
+{{ define "main" }}
+    <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+            {{ template "error-summary" .Error }}
+
+            {{ if .Success }}
+                {{ template "success-banner" "You have taken the investigation off hold." }}
+            {{ end }}
+
+            <h1 class="govuk-heading-l app-!-embedded-hide">Investigation off hold</h1>
+
+            <form class="form" method="POST">
+                <input type="hidden" name="xsrfToken" value="{{ .XSRFToken }}"/>
+                <table class="govuk-table table__no-border">
+                    <tbody class="govuk-table__body">
+                    <tr class="govuk-table__row">
+                        <th scope="row" class="govuk-table__header">Title</th>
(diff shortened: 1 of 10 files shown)


web/template/investigation_on_hold.gohtml
```diff
@@ -52,7 +52,7 @@
                             <div class="govuk-radios__item">
                                 <input class="govuk-radios__input" id="f-reason-2" name="reason" type="radio" value="LA Investigation" {{ if eq "LA Investigation" .Reason }}checked{{ end }}>
                                 <label class="govuk-label govuk-radios__label" for="f-reason-2">
-                                    LA Investigation
+                                    Local Authority Investigation
```
GemTay, line 55: Small content tweaks from speaking to Claire & Katie

hawx: Although this won't match what is displayed for `.HoldPeriod.Reason` on the take investigation off hold page

web/template/investigation_off_hold.gohtml
```diff
@@ -0,0 +1,47 @@
+{{ template "page" . }}
+
+{{ define "title" }}{{ if .Error.Any }}Error: {{ end }}Investigation off hold{{ end }}
```
hawx, line 3: I think these titles are kind of confusing because it sounds like it is describing the current state of the hold (but you don't have to change them). But it did make me wonder whether the on/off hold pages should actually be one page that switches between depending on what the state is?

GemTay: Yeah those titles are confusing, i'll ask Katie and maybe get a ticket created to change them. I agree it would be better to have 1 template with switches based off the `isOnHold` value, however it might be better to do this in a separate ticket as this work is blocking the edit investigation ticket, and its just opened a can of worms in my mind...I think there needs to be more defensive coding, eg checking the state and preventing putting on/off hold if the state is already that which the user is intending. Im thinking ahead to when we don't have angular JS and these actions may have their own page, and the `id` URL param could be changed, or a user could go back to a previously visited page with a different ID, we should check the state and maybe displaying a message to say "This investigation is already on hold" etc (Unlikely scenarios I know, but the fact that you can currently put an investigation on hold which is already on hold etc needs sorting)

hawx: Yeah I was thinking more that than just the template with my comment, so you'd have `/investigation-hold?id=X` (or whatever) and it would display the correct thing.

hawx (approved):

PR #180, opened 2022-11-08
VEGA-1442 recreate edit investigation form

hiboabd:
#minor

### DIFF
--- a/internal/server/edit_investigation.go
+++ b/internal/server/edit_investigation.go
@@ -1,0 +1,71 @@
+package server
+
+import (
+	"github.com/ministryofjustice/opg-go-common/template"
+	"github.com/ministryofjustice/opg-sirius-lpa-frontend/internal/sirius"
+	"net/http"
+	"strconv"
+)
+
+type EditInvestigationClient interface {
+	EditInvestigation(ctx sirius.Context, investigationID int, investigation sirius.Investigation) error
+	Investigation(ctx sirius.Context, id int) (sirius.Investigation, error)
+}
+
+type editInvestigationData struct {
+	XSRFToken            string
+	Success              bool
+	Error                sirius.ValidationError
+	Investigation        sirius.Investigation
+	ApprovalOutcomeTypes []string
+}
+
+var approvalOutcomeTypes = []string{"Court Application", "Further Action", "No Further Action", "Instrument no longer Valid"}
+
+func EditInvestigation(client EditInvestigationClient, tmpl template.Template) Handler {
+	return func(w http.ResponseWriter, r *http.Request) error {
+		investigationID, err := strconv.Atoi(r.FormValue("id"))
+		if err != nil {
+			return err
+		}
+
+		ctx := getContext(r)
+
+		data := editInvestigationData{
+			XSRFToken:            ctx.XSRFToken,
+			ApprovalOutcomeTypes: approvalOutcomeTypes,
+		}
+
+		if r.Method == http.MethodPost {
+			investigation := sirius.Investigation{
+				Title:                    postFormString(r, "title"),
+				Information:              postFormString(r, "information"),
+				Type:                     postFormString(r, "type"),
+				DateReceived:             postFormDateString(r, "dateReceived"),
+				ApprovalDate:             postFormDateString(r, "approvalDate"),
+				ApprovalOutcome:          postFormString(r, "approvalOutcome"),
+				InvestigationClosureDate: postFormDateString(r, "investigationClosureDate"),
+			}
+
+			err = client.EditInvestigation(ctx, investigationID, investigation)
+
+			if ve, ok := err.(sirius.ValidationError); ok {
+				w.WriteHeader(http.StatusBadRequest)
+				data.Error = ve
+				data.Investigation = investigation
+			} else if err != nil {
+				return err
+			} else {
+				data.Success = true
+			}
+		}
+
+		investigation, err := client.Investigation(ctx, investigationID)
+		if err != nil {
+			return err
+		}
+		data.Investigation = investigation
+
+		return tmpl(w, data)
+	}
+}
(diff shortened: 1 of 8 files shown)


hawx (dismissed):

internal/server/edit_investigation.go
```diff
@@ -0,0 +1,72 @@
...
+
+type editInvestigationData struct {
+	XSRFToken            string
+	Success              bool
+	Error                sirius.ValidationError
+	Investigation        sirius.Investigation
+	ApprovalOutcomeTypes []string
+}
+
+var approvalOutcomeTypes = []string{"Court Application", "Further Action", "No Further Action", "Instrument no longer Valid"}
```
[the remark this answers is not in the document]
hiboabd: 😅

internal/sirius/client.go
```diff
@@ -111,6 +112,21 @@ type ValidationError struct {
 	Field  FieldErrors `json:"validation_errors"`
 }
 
+type SiriusFieldErrors map[string][]string
+
+type SiriusValidationError struct {
```
hawx, line 117: I'm confused by this, if it returns the errors as `validation_errors` what was the problem with using `ValidationError`?

hawx: This will be slightly annoying (and hard to explain) so hopefully this example makes some sense, and you can adapt it to what is needed here

```go
type FlexibleFields map[string]json.RawMessage
type StrictFields map[string]map[string]string

func toStrict(x FlexibleFields) StrictFields {
	s := StrictFields{}
	for k, v := range x {
		var asMap map[string]string
		if err := json.Unmarshal(v, &asMap); err == nil {
			s[k] = asMap
			continue
		}

		var asSlice []string
		if err := json.Unmarshal(v, &asSlice); err == nil {
			s[k] = map[string]string{"": strings.Join(asSlice, "")}
			continue
		}

		panic("could not parse error") // but should just make toStrict return an error
	}
	return s
}

func main() {
	var x FlexibleFields
	json.Unmarshal([]byte(`{"field": {"reason": "desc"}}`), &x)
	var y FlexibleFields
	json.Unmarshal([]byte(`{"field": ["desc"]}`), &y)

	fmt.Println(toStrict(x), toStrict(y))
	// => map[field:map[reason:desc]] map[field:map[:desc]]
}
```

hiboabd: 🤦‍♀️ good point, i'll give that a go!

hawx: Another thing (that you can ignore if you want) is that we never use the inner map key, so maybe the best representation for `FieldErrors` would be `map[string][]string` and we should transform the `map[string]map[string]string` version to that instead. That way you wouldn't have to put in the `strings.Join` "hack"

hiboabd: I looked into this but it seems more effort to always need to transform the validation errors returned from `map[string]map[string]string` to `map[string][]string` considering most of the time it will be in that format 

```
	var asFieldErrors map[string]map[string]string
	if err := json.Unmarshal(v, &asFieldErrors); err == nil {
		for key, value := range asFieldErrors {
			for _, value1 := range value {
				s[key] = []string{value1}
			}
		}
		continue
	}
```

internal/sirius/edit_investigation.go
```diff
@@ -24,12 +25,24 @@ func (c *Client) EditInvestigation(ctx Context, investigationID int, investigati
...
+			return err
+		}
+		formattedFieldErrors, err := formatToFieldErrors(x)
+		if err != nil {
 			return err
 		}
-		return FormatToValidationError(v)
+		v.Field = formattedFieldErrors
+		v.Detail = "Payload failed validation"
+		return v
```
hiboabd, line 45: I originally only wanted to unmarshal the response body to `FlexibleFields` if it failed to decode the response body to the `ValidationError` type i.e. in the case that the error was coming from the handler. However, i kept getting an EOF error which i couldn't figure out..

```go
if resp.StatusCode == http.StatusBadRequest {
	var v ValidationError
	if err := json.NewDecoder(resp.Body).Decode(&v); err != nil {
		var x FlexibleFields
		if err := json.Unmarshal(body, &x); err != nil {
			return err
		}
		formattedFieldErrors, err := toStrict(x)
		if err != nil {
			return err
		}
		v.Field = formattedFieldErrors
		v.Detail = "Payload failed validation"
		return v
	}
	return v
}
```

hawx: I think you can simplify this by using `FlexibleField` to replace `FieldErrors`, then here you could have:

```go
// removed the io.ReadAll here

if resp.StatusCode == http.StatusBadRequest {
	var v struct {
		Detail string              `json:"detail"`
		Field  flexibleFieldErrors `json:"validation_errors"`
	}
	if err := json.NewDecoder(resp.Body).Decode(&v); err != nil {
		return err
	}
	fieldErrors, err := v.Field.toFieldErrors()
	if err != nil {
		return err
	}
	return ValidationError{
		Detail: v.Detail,
		Field:  fieldErrors,
	}
}
```

I renamed `FlexibleFields` to `flexibleFieldErrors`, and then I've made the transformation function a method on that type, like:

```go
type flexibleFieldErrors map[string]json.RawMessage

func (f flexibleFieldErrors) toFieldErrors() (FieldErrors, error) {
	s := FieldErrors{}

	for k, v := range f {
		var asSlice []string
		if err := json.Unmarshal(v, &asSlice); err == nil {
			s[k] = map[string]string{"": strings.Join(asSlice, "")}
			continue
		}

		var asMap map[string]string
		if err := json.Unmarshal(v, &asMap); err == nil {
			s[k] = asMap
			continue
		}

		return nil, errors.New("could not parse field validation_errors")
	}

	return s, nil
}
```

and then you don't need to check for the `validation_errors` key.

internal/sirius/client.go
hawx: Yeah that makes sense

internal/sirius/edit_investigation.go
hiboabd: That's much cleaner, thanks i've made that change

hawx (approved):

PR #211, opened 2022-12-05
VEGA-1370 recreate search results page

GemTay:
VEGA 1370


web/template/layout/header.gohtml
```diff
@@ -17,6 +17,23 @@
...
             </li>
+            <li class="moj-header__navigation-item">
+              <div class="moj-search moj-search--ondark moj-search--inline">
+                <form class="form" method="get" data-module="search">
+                  <div class="govuk-form-group">
+                    <label class="govuk-label moj-search__label govuk-visually-hidden" for="search">
+                      Search
+                    </label>
+                    <input class="govuk-input moj-search__input" id="search" name="search" type="search">
+                  </div>
```
gregtyler, line 28: If you set the `action` and `name` attributes, the form will automatically submit to `/search?term={term}` and you won't need the JavaScript to do it:
```suggestion
                <form class="form" method="get" action="{{ url "/search" }}" data-module="search">
                  <div class="govuk-form-group">
                    <label class="govuk-label moj-search__label govuk-visually-hidden" for="search">
                      Search
                    </label>
                    <input class="govuk-input moj-search__input" id="search" name="term" type="search">
                  </div>
```

web/template/layout/pagination.gohtml
```diff
@@ -0,0 +1,35 @@
...
+                <li class="moj-pagination__item  moj-pagination__item--next">
+                    <a class="moj-pagination__link" href="{{ .SearchTerm }}{{ .Filters }}&page={{ .NextPage }}">Next<span class="govuk-visually-hidden"> page</span></a>
+                </li>
+            {{ end }}
+        </ul>
+    {{ end }}
+{{ end }}
+
+{{ define "pagination-footer" }}
+    <nav class="moj-pagination" aria-label="Pagination navigation (in footer)">
```
gregtyler, line 32: I don't think "in footer" is needed her

internal/server/server.go
```diff
@@ -208,3 +210,17 @@ func translateRefData(types []sirius.RefDataItem, tmplHandle string) string {
...
+		return 1
+	}
+
+	v, err := strconv.Atoi(page)
+	if err != nil {
+		return 1
+	}
+
+	return v
+}
```
gregtyler, line 226: Probably best to put this in `internal/server/search.go` so it's clearer what it's for

internal/server/search.go
```diff
@@ -0,0 +1,107 @@
...
+
+		results, pagination, err := client.Search(ctx, searchTerm, getPage(r), filters.PersonType)
+		if err != nil {
+			return err
+		}
+
+		if results.Total.Count == 0 {
+			re := regexp.MustCompile(`\D+`)
+			input := re.ReplaceAllString(searchTerm, "")
+			isUid, err := regexp.MatchString(`^\d{12}$`, re.ReplaceAllString(input, ""))
```
gregtyler, line 87: Looks like you're removing non-numeric characters twice:
```suggestion
			input := re.ReplaceAllString(searchTerm, "")
			isUid, err := regexp.MatchString(`^\d{12}$`, input)
```

web/template/layout/header.gohtml
gregtyler: This will also work better when embedded, since it needs to include the prefix

gregtyler (commented):
Worth noting that this only shows one row per donor. For most fields that won't matter, but it means you only see one type/status which could be confusing if you're looking for the other

web/template/layout/header.gohtml
GemTay: This works well however the JavaScript is needed to populate the search input on submit, so that you don't lose what you entered in the box, allowing users to edit it.

GemTay:
> Worth noting that this only shows one row per donor. For most fields that won't matter, but it means you only see one type/status which could be confusing if you're looking for the other

@gregtyler Is this different to linked actors? I'll have a look into this

web/template/layout/header.gohtml
gregtyler: You can keep the `window.addEventListener('load')` for that, but you shouldn't need to intercept the default submit event. I don't think it adds anything and it means the form currently doesn't work without JavaScript.

PR #232, opened 2022-12-08
VEGA-1239 recreate create document form (create draft)

hiboabd:


web/template/create_document.gohtml
```diff
@@ -0,0 +1,207 @@
+{{ template "page" . }}
+
+{{ define "title" }}{{ if .Error.Any }}Error: {{ end }}Create Document{{ end }}
+
+{{ define "main" }}
```
hiboabd, line 5: As design may change, I will split this into smaller templates once the design is confirmed

web/template/create_document.gohtml
```diff
@@ -0,0 +1,207 @@
...
+                                    {{ if eq (len .Recipients) 0 }}
+                                        <p class="govuk-body">No recipients available to select</p>
+                                    {{ else }}
+                                    <div class="govuk-!-width-one-half govuk-form-group {{ if .Error.Field.selectRecipient }}govuk-form-group--error{{ end }}">
+                                        <div class="govuk-form-group {{ if .Error.Field.selectRecipient }}govuk-form-group--error{{ end }}">
+                                            <label class="govuk-label" for="f-selectRecipient">
+                                                Select recipient
+                                            </label>
+                                            {{ template "errors" .Error.Field.selectRecipient }}
+                                            <select class="govuk-select {{ if .Error.Field.selectRecipient }}govuk-select--error{{ end }}" id="f-selectRecipient" name="selectRecipient">
```
gregtyler, line 50: In the current design you can select multiple recipients, though I'm not sure if that's used particularly regularly

internal/sirius/document.go
```diff
@@ -0,0 +1,121 @@
...
+
+type documentTemplateApiResponse map[string]json.RawMessage
+type insertApiResponse map[string]json.RawMessage
+
+type UniversalTemplateData struct {
+	Location        string `json:"location"`
+	OnScreenSummary string `json:"onScreenSummary"`
+}
+
+type DocumentTemplateApiData struct {
```
gregtyler, line 27: I don't think this is ever used externally, so it can be excluded from export to avoid confusion:
```suggestion
type documentTemplateApiData struct {
```

internal/sirius/document.go
```diff
@@ -0,0 +1,121 @@
...
+
+func (d documentTemplateApiResponse) toDocumentData() ([]DocumentTemplateData, error) {
+	var s []DocumentTemplateData
+
+	for k, v := range d {
+		var asDocumentTemplateData DocumentTemplateData
+		asDocumentTemplateData.TemplateId = k
+
+		var asDocumentTemplateApiData DocumentTemplateApiData
+		if err := json.Unmarshal(v, &asDocumentTemplateApiData); err == nil {
```
gregtyler, line 52: It seems weird to be unmarshalling here having just done so in `DocumentTemplates`. Could you not do:
```go
type documentTemplateApiResponse map[string]DocumentTemplateApiData
```

internal/sirius/edit_case_test.go
```diff
@@ -34,6 +34,7 @@ func TestEditCase(t *testing.T) {
 						Path:   dsl.String("/lpa-api/v1/lpas/800"),
 						Body: map[string]interface{}{
 							"status": "Suspended",
+							"donor":  Person{ID: 1},
```
gregtyler, line 37: I don't think this should be here: you can't change the donor when editing an LPA, it'll just be ignored by the API

internal/sirius/document.go
hiboabd: The issue is with the inserts field which will either return an empty array `[]` or a `map[string]json.RawMessage`. So it will throw an error when trying to decode the insert field when it's an empty array

internal/sirius/edit_case_test.go
hiboabd: It doesn't seem to be ignored, at least when i run the tests locally

`Error: Expected nil, but got: &errors.errorString{s:"Actual interactions do not match expected interactions for mock MockService.\n\nIncorrect requests:\n\tPUT /lpa-api/v1/lpas/800 (request body did not match)\n\nDiff with interaction: \"A request to edit the LPA\" given \"I have a pending case assigned\"... Could not find key \"donor\" (keys present are: status) at $.body\n\nSee /Users/hibo.abdilaahi/Documents/GitHub/opg-sirius-lpa-frontend/logs/pact.log for details.\n"}`

The solution seems to be to cast the donor field to a pointer type `*Person`

internal/server/create_document.go
```diff
@@ -0,0 +1,293 @@
...
+					Town:                  postFormString(r, "town"),
+					County:                postFormString(r, "county"),
+					Postcode:              postFormString(r, "postcode"),
+					IsAirmailRequired:     postFormString(r, "isAirmailRequired") == "Yes",
+					PhoneNumber:           postFormString(r, "phoneNumber"),
+					Email:                 postFormString(r, "email"),
+					CorrespondenceByPost:  postFormCheckboxChecked(r, "correspondenceBy", "post"),
+					CorrespondenceByEmail: postFormCheckboxChecked(r, "correspondenceBy", "email"),
+					CorrespondenceByPhone: postFormCheckboxChecked(r, "correspondenceBy", "phone"),
+					CorrespondenceByWelsh: postFormCheckboxChecked(r, "correspondenceBy", "welsh"),
```
hawx, line 175: I know this is matching what we have, but I think these aren't needed because they won't affect the letter (and non-case contacts aren't used for anything else)

internal/server/create_document.go
```diff
@@ -0,0 +1,293 @@
...
+	var documentTemplateTypes []sirius.RefDataItem
+	for _, dt := range documentTemplateData {
+		for _, refData := range documentTemplateRefData {
+			if refData.Handle == dt.OnScreenSummary {
+				translatedRefDataItem := sirius.RefDataItem{
+					Handle:         dt.TemplateId,
+					Label:          refData.Label,
+					UserSelectable: false,
+				}
+				documentTemplateTypes = append(documentTemplateTypes, translatedRefDataItem)
```
hawx, line 256: could `break` after this to save going through the whole `documentTemplateRefData` list

internal/server/create_document.go
```diff
@@ -0,0 +1,293 @@
...
+	var documentTemplateInserts []InsertDisplayData
+	for _, in := range selectedTemplateInserts {
+		for _, refData := range documentTemplateRefData {
+			if refData.Handle == in.OnScreenSummary {
+				translatedRefDataItem := InsertDisplayData{
+					Handle: in.InsertId,
+					Label:  refData.Label,
+					Key:    in.Key,
+				}
+				documentTemplateInserts = append(documentTemplateInserts, translatedRefDataItem)
```
hawx, line 274: same as above, could `break` here

internal/sirius/case_type.go
```diff
@@ -11,7 +11,7 @@ const (
 
 func ParseCaseType(s string) (CaseType, error) {
 	switch s {
-	case "lpa":
+	case "lpa", "LPA":
```
hawx, line 14: do you need this, and if so will you need `EPA` for the other?

internal/sirius/document_test.go
```diff
@@ -0,0 +1,96 @@
...
+				if tc.expectedError == nil {
+					assert.Nil(t, err)
+				} else {
+					assert.Equal(t, tc.expectedError(pact.Server.Port), err)
+				}
+				return nil
+			}))
+		})
+	}
+}
```
hawx, line 96: it would be good to have cases for the different ways the data can be returned to show the deserialising works

internal/server/create_document.go
hiboabd: I see what you are saying but i'm a bit hesitant to change how the form currently works. I can check with katie and see what she thinks

internal/sirius/case_type.go
hiboabd: Yeah the api request returns the case type in upper case form, it will be needed for epa

internal/sirius/document_test.go
```diff
@@ -72,6 +72,40 @@ func TestDocumentTypes(t *testing.T) {
...
+						Method: http.MethodGet,
+						Path:   dsl.String("/lpa-api/v1/templates/lpa"),
+					}).
+					WillRespondWith(dsl.Response{
+						Status: http.StatusOK,
+						Body: dsl.Like(map[string]interface{}{
+							"DD": dsl.Like(map[string]interface{}{
+								"onScreenSummary": dsl.Like("DDONSCREENSUMMARY"),
+								"location":        dsl.Like(`lpa\/DD.html.twig`),
+								"inserts":         dsl.Like(map[string]interface{}{}),
```
hawx, line 92: since the API returns `[]` when empty you'll want this to be `[]interface{}{}`

internal/sirius/document.go
```diff
@@ -0,0 +1,121 @@
...
+	Type            string `json:"type"`
+	SystemType      string `json:"systemType"`
+	FileName        string `json:"fileName,omitempty"`
+	Content         string `json:"content"`
+}
+
+type documentTemplateApiResponse map[string]json.RawMessage
+type insertApiResponse map[string]json.RawMessage
+
+type UniversalTemplateData struct {
```
hawx, line 22: sorry I didn't pick up on this before, but I think repeating these two fields in each struct makes more sense than embedding this struct within them

hiboabd: I've kept the `UniversalTemplateData` for the deserialisation in `toDocumentData` and `toInsertData` but i have repeated the fields for the other structs

hawx (approved):

gregtyler (approved):

PR #240, opened 2022-12-21
VEGA-1239 frontend changes to create document form

hiboabd:
Makes changes to create document form following design changes
Design: https://www.figma.com/file/2G9pvk1ET4KHlYN8LWbBrg/Sirius-LPA-2022-Create-doc-%26-edit-draft-widget?node-id=1%3A48&t=OknGqCPPTZeo6BLI-0

#minor

[LPA-2022]: https://opgtransform.atlassian.net/browse/LPA-2022?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


web/assets/toggle-tabs.js
```diff
@@ -0,0 +1,37 @@
+const toggleTabs = () => {
```
hiboabd, line 1: Not sure if toggle is the right term for the action...

web/assets/toggle-insert-checkboxes.js
```diff
@@ -0,0 +1,23 @@
+const toggleInsertCheckboxes = () => {
+    /*Insert checkboxes may appear twice and therefore js is needed to make sure
+    that a check in one tab is displayed in the other*/
+
```
hiboabd, line 4: A visual illustration of what this is solving, in case this comment doesn't make sense 
Screenshot 2022-12-21 at 13 38 58
Screenshot 2022-12-21 at 13 39 09

hiboabd: I tried to look displaying the same checkbox twice but as tab content sits in different containers to one another, they need to be separate checkboxes

internal/server/create_document.go
```diff
@@ -277,9 +313,43 @@ func translateInsertData(selectedTemplateInserts []sirius.Insert, documentTempla
 			}
 		}
 	}
+
+	sort.Slice(documentTemplateInserts, func(i, j int) bool {
```
hiboabd, line 317: Sorting in alphabetical order, could probably be a helper function

internal/server/create_document.go
```diff
@@ -97,46 +101,80 @@ func CreateDocument(client CreateDocumentClient, tmpl template.Template) Handler
...
+
 			if data.TemplateSelected.TemplateId != "" {
 				data.DocumentInsertTypes = translateInsertData(data.TemplateSelected.Inserts, data.DocumentTemplateRefData)
+				data.DocumentInsertKeys = getSortedInsertKeys(data.TemplateSelected.Inserts)
 			}
 
 			hasViewedInsertPage := r.FormValue("hasViewedInserts")
-			if hasViewedInsertPage == "true" {
+			hasNoInsertsToSelect := data.TemplateSelected.TemplateId != "" && len(data.DocumentInsertTypes) == 0
+			uniqueInserts := removeDuplicateStr(r.Form["insert"])
```
hiboabd, line 129: Had to handle remove duplicate inserts as an insert can appear in multiple tabs and all ids are posted with the form

internal/server/create_document.go
```diff
@@ -3,8 +3,10 @@ package server
 import (
 	"github.com/ministryofjustice/opg-go-common/template"
 	"github.com/ministryofjustice/opg-sirius-lpa-frontend/internal/sirius"
+	"golang.org/x/exp/slices"
```
hiboabd, line 6: I thought the contains function below was handy and saved me writing a function to do it but don't know if requiring an extra package would have an affect on performance..

internal/server/create_document.go
```diff
@@ -277,9 +313,43 @@ func translateInsertData(selectedTemplateInserts []sirius.Insert, documentTempla
...
+	for _, in := range selectedTemplateInserts {
+		if !slices.Contains(documentInsertKeys, in.Key) {
+			documentInsertKeys = append(documentInsertKeys, in.Key)
+		}
+	}
+	// insert api response infrequently includes the key all
+	// note api keys are lowercase so this check is case-sensitive
+	if !slices.Contains(documentInsertKeys, "all") {
+		documentInsertKeys = append(documentInsertKeys, "all")
+	}
```
hiboabd, line 335: I'll create a ticket to clean up the document template api response so it's more consistent with the inclusion of the all key and then this can be removed? Or conversely we could remove it from the api response and always handle it in the frontend

internal/server/create_document.go
```diff
@@ -277,9 +313,43 @@ func translateInsertData(selectedTemplateInserts []sirius.Insert, documentTempla
...
+
+	sort.Strings(documentInsertKeys)
+	return documentInsertKeys
+}
+
+func removeDuplicateStr(strSlice []string) []string {
+	allKeys := make(map[string]bool)
```
hawx, line 342: a "fun" trick is you can use `map[string]struct{}` and insert `struct{}{}` values where you don't care about the value, these are empty values so reduces memory use by a byte per item... much performance

web/template/create_document.gohtml
```diff
@@ -5,204 +5,155 @@
...
-                                                        Welsh
-                                                    </label>
-                                                </div>
+            {{ if .HasSelectedAddNewRecipient }}
+                <form class="form" method="GET">
+                    {{ $url := (printf "create-document?id=%d&case=%s&templateId=%s&hasViewedInserts=true" .Case.ID .Case.CaseType .TemplateSelected.TemplateId) }}
+                    {{ range .SelectedInserts }}
+                        {{ $url = (printf "%s&insert=%s" $url .)  }}
+                    {{ end }}
+                    <a href="{{$url}}" class="govuk-back-link">Back</a>
```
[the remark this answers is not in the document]
hawx: couldn't you do

```
{{ range .SelectedInserts }}
  <input type="hidden" name="selectedInserts" value="{{ . }}" />
{{ end }}
```

to use hidden inputs for these?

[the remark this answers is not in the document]
hawx: not that I'm totally against keeping the state in the URL, for your template function you probably needed to return `template.HTML("whatever string")` instead of a string, then it won't escape it

web/template/create_document.gohtml
```diff
@@ -5,204 +5,155 @@
...
-                                                        Phone
-                                                    </label>
-                                                </div>
-                                                <div class="govuk-checkboxes__item">
-                                                    <input class="govuk-checkboxes__input" id="f-correspondenceBy-welsh" name="correspondenceBy" type="checkbox" value="welsh">
-                                                    <label class="govuk-label govuk-checkboxes__label" for="f-correspondenceBy-welsh">
-                                                        Welsh
-                                                    </label>
-                                                </div>
+            {{ if .HasSelectedAddNewRecipient }}
```
hawx, line 15: it might be nice to pull out bits into templates here, you can keep them in the same file and have

```
{{ define "addNewRecipient" }}
  <!-- this content -->
{{ end }}
```

then put `{{ template "addNewRecipient" . }}` like you would for any other

web/template/layout/add-recipient-form.gohtml
```diff
@@ -0,0 +1,68 @@
+{{ define "add-recipient" }}
```
hawx, line 1: since this will only ever be used by one template it would be better in the same file

web/template/layout/case-details.gohtml
```diff
@@ -0,0 +1,3 @@
+{{ define "case-details" }}
+    <p class="govuk-body"><strong>{{ if eq .Case.SubType "pfa" }}PFA{{ else if eq .Case.SubType "lpa" }}LPA{{ end }} {{ .Case.UID }}</strong></p>
```
hawx, line 2: I'm not sure when this slipped in but this should surely be saying PFA or HW?

web/template/layout/insert-tabs.gohtml
```diff
@@ -0,0 +1,64 @@
...
+                <table class="govuk-table my-table table__no-border">
+                    <tbody class="govuk-table__body">
+                    {{ range $t := $.DocumentInsertTypes }}
+                        {{if eq $t.Key $k }}
+                            <tr class="govuk-table__row">
+                                <td class="govuk-table__cell">
+                                    <div class="govuk-checkboxes" data-module="govuk-checkboxes">
+                                        <div class="govuk-checkboxes__item">
+                                            <input class="govuk-checkboxes__input"
+                                                   id="f-{{ $t.Handle }}"
```
hawx, line 25: because you can have the same handle in multiple tabs you can end up with an id being reused on the same page, could these be prefixed with the tab too?

hawx (commented):
it would be cool to remove the extra line at the end of the inserts

image

web/template/layout/case-details.gohtml
hiboabd: As far as i can tell it has always been pfa or lpa... i can double check when everyone is back

web/template/create_document.gohtml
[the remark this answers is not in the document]
hiboabd: Using hidden inputs isn't an option as it's a link element. I've given returning a` template.HTML` string a go and it still escapes the `=` and `&` so for now i have kept this as is

hawx (approved):

PR #244, opened 2023-01-06
Respond appropriately to bad search requests

gregtyler:
Two main changes here:
- Respond to bad search requests with a ValidationError, not a plain text error
- If the request accepts JSON, return a RFC7807 error page rather than an HTML page

I also added coverage for fieldset labels in dark mode

#patch

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -1,6 +1,7 @@
 package server
 
 import (
+	"encoding/json"
 	"net/http"
 	"net/url"
 	"strconv"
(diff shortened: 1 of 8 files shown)


internal/server/server.go
```diff
@@ -164,6 +171,30 @@ func errorHandler(logger Logger, tmplError template.Template, prefix, siriusURL
 					correlationId = statusError.CorrelationId
 				}
 
+				if strings.Contains(r.Header.Get("Accept"), "application/json") {
```
hawx, line 174: This could be wrong, if I sent up `Accept: text/html, application/json;q=0.8` I'd get JSON but want HTML. Having said that, I'm not sure if we care?

internal/server/server.go
```diff
@@ -141,6 +142,12 @@ func (e RedirectError) To() string {
 	return string(e)
 }
 
+type RfcError struct {
```
hawx, line 145: Would `ProblemError` be a better name, what RFC does it refer to?

hawx (approved):

internal/server/server.go
gregtyler: Personally I don't think we care enough to do full parsing, but I've simplified to an exact match to avoid that case (and since we're in full control of what header we send).

internal/server/server.go
gregtyler: Yeah, that's clearer and updated now. The RFC in question is [7807](https://www.rfc-editor.org/rfc/rfc7807) FWIW

PR #250, opened 2023-01-11
VEGA-1576-cancelled-requests-bug

GemTay:
Changed error code to 499 for cancelled requests (commonly caused autocomplete search input fields), so that we can distinguish these separately from true 500 errors

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -1,7 +1,9 @@
 package server
 
 import (
+	"context"
 	"encoding/json"
+	"errors"
 	"net/http"
 	"net/url"
 	"strconv"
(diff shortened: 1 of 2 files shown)


internal/server/server.go
```diff
@@ -195,6 +197,10 @@ func errorHandler(logger Logger, tmplError template.Template, prefix, siriusURL
 					return
 				}
 
+				if errors.Is(err, context.Canceled) {
+					code = 499
+				}
+
```
hiboabd, line 203: Does this need to be tested?

gregtyler (approved):

PR #264, opened 2023-01-20
VEGA-1239 design changes to document forms

hiboabd:


web/assets/main.scss
```diff
@@ -77,3 +77,7 @@ $govuk-assets-path: "/lpa/frontend/assets/";
         border-bottom: none;
     }
 }
+
+.app-\!-select-draft-width {
+  width: 88%;
+}
```
hiboabd, line 83: Not ideal but the grid classes available to set the width don't go wide enough (the widest class `govuk-grid-column-three-quarters` only sets the width to 75%).
Screenshot 2023-01-20 at 15 49 41

gregtyler: Could you just go full width with `govuk-grid-column-full`?

web/template/create_document.gohtml
```diff
@@ -38,12 +38,11 @@
 
             {{ else if .TemplateSelected.TemplateId }}
                 <a href="create-document?id={{ .Case.ID }}&case={{ .Case.CaseType }}" class="govuk-back-link">Back</a>
-
                 {{ template "case-details" . }}
                 {{ template "select-insert" . }}
 
             {{ else }}
-                {{ template "case-details" . }}
+                <p class="govuk-body govuk-!-margin-top-8"><strong>{{ if eq .Case.SubType "pfa" }}PFA{{ else if eq .Case.SubType "lpa" }}LPA{{ end }} {{ .Case.UID }}</strong></p>
```
gregtyler, line 45: To avoid duplicating the `case-details` partial, you could just wrap in a `div` to add the margin:
```suggestion
                <div class="govuk-!-margin-top-8">{{ template "case-details" . }}</div>
```

web/template/create_document.gohtml
```diff
@@ -121,7 +120,11 @@
                 <div class="govuk-button-group govuk-!-padding-top-5">
                     <button class="govuk-button" data-module="govuk-button" type="submit" name="recipientControls" value="addNewRecipient">Continue</button>
                 </div>
-                <a data-app-iframe-cancel class="govuk-link govuk-link--no-visited-state" href="#">Cancel</a>
+                {{ $url := (printf "create-document?id=%d&case=%s&templateId=%s&hasViewedInserts=true" .Case.ID .Case.CaseType .TemplateSelected.TemplateId) }}
+                {{ range .SelectedInserts }}
+                    {{ $url = (printf "%s&insert=%s" $url .)  }}
+                {{ end }}
+                <a href="{{$url}}" class="govuk-link govuk-link--no-visited-state">Cancel</a>
```
gregtyler, line 127: You might want to consider adding a `Back` property to `createDocumentData` and do this logic in Go rather than the templating language. Either way you also need to apply the prefix to it so it'll always link to the right place.

web/template/edit_document.gohtml
```diff
@@ -9,19 +9,20 @@
...
                 <meta data-app-reload="reload-timeline"/>
+                {{ template "success-banner" "Draft document published" }}
             {{ else if .SaveAndExit }}
                 <meta data-app-reload="saveAndExit"/>
             {{ else if .PreviewDraft }}
                 <a href="{{ sirius (printf "/lpa-api/v1/documents/%s/download" .DownloadUUID) }}" target="_blank"
                    data-module="app-auto-click">Download preview document</a>
             {{ end }}
 
+            <a href="create-document?id={{ .Case.ID }}&case={{ .Case.CaseType }}" class="govuk-back-link">Back</a>
```
gregtyler, line 20: This also should be prefixed:
```suggestion
            <a href="{{ prefix (printf "/create-document?id=%d&case=%s" .Case.ID .Case.CaseType) }} class="govuk-back-link">Back</a>
```

web/assets/main.scss
hiboabd: To get back to your comment below @gregtyler , `govuk-grid-column-full` doesn't work as i don't want the drop down to span the full width as it knocks the select button to sit below it. The next widest grid class `govuk-grid-column-three-quarters` is not wide enough.

gregtyler: Ah sorry, I misunderstood what this was applying to. In that case it'd probably be safer to use flexbox, but this setup will work fine.

web/template/create_document.gohtml
```diff
@@ -121,7 +106,11 @@
...
```
gregtyler, line 113: Can this use `.Back` too?
```suggestion
                <a href="{{ prefix .Back }}" class="govuk-link govuk-link--no-visited-state">Cancel</a>
```

web/assets/main.scss
hiboabd: Switched to flexbox, as it was cleaner

gregtyler (approved):

PR #265, opened 2023-01-24
Add new fields to Add/Resolve Complaint forms

gregtyler:
Add ComplainantCategory and Origin

Fixes VEGA-1654 #minor

### DIFF
--- a/internal/server/add_complaint.go
+++ b/internal/server/add_complaint.go
@@ -8,11 +8,13 @@ import (
 
 	"github.com/ministryofjustice/opg-go-common/template"
 	"github.com/ministryofjustice/opg-sirius-lpa-frontend/internal/sirius"
+	"golang.org/x/sync/errgroup"
 )
 
 type AddComplaintClient interface {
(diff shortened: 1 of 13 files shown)


internal/server/add_complaint.go
```diff
@@ -107,8 +109,11 @@ type addComplaintData struct {
 	Success   bool
 	Error     sirius.ValidationError
 
-	Categories map[string]complaintCategory
-	Complaint  sirius.Complaint
+	Categories            map[string]complaintCategory
```
GemTay, line 112: Not required for this ticket but im curious as to why we hardcode the categories and dont fetch them from the API instead :raised_eyebrow:

gregtyler: They're not actually stored in the API, only the frontend! If you manipulate the request, there's nothing stopping you setting the category/subcategory to any string you like. [VEGA-1667](https://opgtransform.atlassian.net/browse/VEGA-1667) should address this.

GemTay (approved):

PR #279, opened 2023-02-15
VEGA-1669 edit complaints form validation

hiboabd:


internal/server/edit_complaint.go
```diff
@@ -64,23 +67,36 @@ func EditComplaint(client EditComplaintClient, tmpl template.Template) Handler {
...
+					data.Complaint = populateComplaint(r)
+					return tmpl(w, data)
+				}
+			}
+
+			category := postFormString(r, "category")
+			if category != "" {
+				if getValidSubcategory(postFormString(r, "category"), r.PostForm["subCategory"]) == "" {
+					return getSubcategoryValidationError(w, r, tmpl, data)
+				}
```
GemTay, line 95: This is all fine but ideally we could just send the data and let the API handle the validation. Im wondering if we could create a custom validator in sirius API that would be able to handle all of this logic to save us having to do it in the frontend. I'm going to have a look into seeing if its a possibility to check the resolution state and conditionally apply validation off that

hiboabd: I think this is possible in the `UpdateComplaintHandler` i'll give it a go

gregtyler: Agree that this should be in the API so it can't be bypassed. If conditional validation is a PITA you could have two validators and switch them at the controller:
```php
$validator = $data['resolutionDate'] ? ResolveComplaint::class : UpdateComplaint::class;

$complaint = $this->commandBus($this->validatePayload($validator, $data));
```
I think we've done this somewhere already but I can't remember where.

hiboabd:
Validation moved to backend so pr can be closed

PR #280, opened 2023-02-21
VEGA-1669 remove unrequired payload failed validation message

hiboabd:
Removes payload failed validation message that appears on validation messages as it is not necessary for users

### DIFF
--- a/web/template/layout/error-summary.gohtml
+++ b/web/template/layout/error-summary.gohtml
@@ -6,7 +6,7 @@
       </h2>
       <div class="govuk-error-summary__body">
         <ul class="govuk-list govuk-error-summary__list">
-          {{ if .Detail }}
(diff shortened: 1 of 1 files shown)


web/template/layout/error-summary.gohtml
```diff
@@ -6,7 +6,7 @@
       </h2>
       <div class="govuk-error-summary__body">
         <ul class="govuk-list govuk-error-summary__list">
-          {{ if .Detail }}
+          {{ if (and .Detail (ne .Detail "Payload failed validation") )}}
```
hiboabd, line 9: I opted to leave Detail property as I think it's required for an error message for the create relationship form 
https://github.com/ministryofjustice/opg-sirius-lpa-frontend/pull/45

gregtyler (approved):

PR #281, opened 2023-02-21
VEGA-1704 combined search result changes

GemTay:
VEGA-1704 #minor

### DIFF
--- a/web/assets/main.scss
+++ b/web/assets/main.scss
@@ -64,11 +64,6 @@ $govuk-assets-path: "/lpa/frontend/assets/";
     width: fit-content;
 }
 
-.app-\!-search-results-table th,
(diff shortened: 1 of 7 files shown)


web/assets/main.scss
```diff
@@ -63,11 +63,6 @@ $govuk-assets-path: "/lpa/frontend/assets/";
     width: fit-content;
 }
 
-.app-\!-search-results-table th,
-.app-\!-search-results-table th button {
-    color: govuk-colour("blue");
```
GemTay, line 68: since we're temporarily removing the sorting functionality on the headings, I have removed this so that the headings are now black, otherwise users could get confused thinking they are a clickable link (verified by claire)

gregtyler (approved):

PR #289, opened 2023-02-28
VEGA-1712 remove inactive attorneys from letter recipient list

GemTay:
VEGA-1712 #patch

### DIFF
--- a/internal/sirius/person.go
+++ b/internal/sirius/person.go
@@ -36,6 +36,7 @@ type Person struct {
 	CompanyReference      string     `json:"companyReference"`
 	PersonType            string     `json:"personType,omitempty"`
 	Cases                 []*Case    `json:"cases,omitempty"`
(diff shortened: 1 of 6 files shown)


internal/sirius/person.go
```diff
@@ -36,6 +36,7 @@ type Person struct {
 	CompanyReference      string     `json:"companyReference"`
 	PersonType            string     `json:"personType,omitempty"`
 	Cases                 []*Case    `json:"cases,omitempty"`
+	SystemStatus          *bool      `json:"systemStatus,omitempty"`
```
GemTay, line 39: ive used `*bool` to be able to make a distinction between a non-existent boolean and false boolean

gregtyler: Do we need to make that distinction? It seems like we treat `false` and `nil` the same

gregtyler (approved):

PR #290, opened 2023-03-02
VEGA-1664 get draft documents via query param

hiboabd:
Can now filter for doc types when retrieving documents by case so client side logic has been removed

### DIFF
--- a/internal/sirius/document.go
+++ b/internal/sirius/document.go
@@ -30,7 +30,7 @@ func (c *Client) Documents(ctx Context, caseType CaseType, caseId int, filterByD
 	var d []Document
 
 	url := fmt.Sprintf("/lpa-api/v1/%s/%d/documents", caseType+"s", caseId)
-	if filterByDocType == true {
+	if filterByDocType {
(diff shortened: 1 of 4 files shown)


internal/sirius/document.go
```diff
@@ -26,10 +26,15 @@ const (
 	TypeSave    string = "Save"
 )
 
-func (c *Client) Documents(ctx Context, caseType CaseType, caseId int) ([]Document, error) {
+func (c *Client) Documents(ctx Context, caseType CaseType, caseId int, filterByDocType bool) ([]Document, error) {
```
gregtyler, line 29: I wasn't initially clear what this variable did from the name. Could we change it to something like "onlyDrafts"? Or make it a string so downstream calls pass `sirius.TypeDraft` in?

hiboabd: Oops initially I was trying to keep it flexible so other document types can be used but that's kind of pointless as only the draft doc type is added to the url in the method.. I'll change it to a string and leave that to be changed if it's needed in the future

gregtyler (approved):

PR #295, opened 2023-03-10
Show inserts when selecting document template

gregtyler:
Adds a component to the template selection screen which shows the available inserts when you select a document. It has the same design and functionality of the existing insert screen but is inline.

I had to create a new data object in a structure that's useful for JavaScript. The component then generates the DOM either from templates or from directly creating each element (using a helper function called createElement).

The component contains a hidden input called `hasViewedInserts` which causes the next page to be skipped. But if the component fails to load then the input isn't added and the user is taken to the fallback page.

This also ensures the documents are sorted by newest-first on the edit screen, to ensure the newly-created document is always shown first.

Fixes VEGA-1751, VEGA-1759 #minor

### DIFF
--- a/internal/server/create_document_test.go
+++ b/internal/server/create_document_test.go
@@ -87,7 +87,8 @@ func TestGetCreateDocument(t *testing.T) {
 					Case:                    caseItem,
 					DocumentTemplateRefData: documentTemplates,
 					DocumentTemplates:       documentTemplateData,
-					Back:                    "/create-document?id=0&case=" + caseType,
+					ComponentDocumentData: buildComponentDocumentData(documentTemplateData, documentTemplates),
+					Back:                  "/create-document?id=0&case=" + caseType,
 				}).
 				Return(nil)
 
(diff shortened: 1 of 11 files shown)


internal/server/create_document_test.go
```diff
@@ -87,7 +87,19 @@ func TestGetCreateDocument(t *testing.T) {
...
+					// 	Templates: []ComponentTemplateData{
+					// 		{
+					// 			Id:      "DD",
+					// 			Inserts: map[string][]ComponentTemplateInsertData{},
+					// 		},
+					// 	},
+					// 	Translations: map[string]string{
+					// 		"DD": "DD Template Label",
+					// 	},
+					// },
```
GemTay, line 100: Think you've accidentally left some commented code here :)

GemTay (dismissed):

internal/server/edit_document.go
```diff
@@ -66,6 +68,11 @@ func EditDocument(client EditDocumentClient, tmpl template.Template) Handler {
 				if err != nil {
 					return err
 				}
+
+				sort.Slice(documents, func(i, j int) bool {
+					return documents[i].ID > documents[j].ID
+				})
```
GemTay, line 74: Would it be worth moving this sorting logic inside the `Documents(...)` function so theyre always sorted? Or is there ever a situation where we wouldn't want them sorted?

PR #298, opened 2023-03-27
VEGA-1717 order recipients in create document

GemTay:
recipients now follow order in ticket. I have also refactored the create document code to remove unnecessary API calls to get each recipient, as they have already been fetched with the case data.
VEGA-1717 #patch

### DIFF
--- a/internal/sirius/person.go
+++ b/internal/sirius/person.go
@@ -62,13 +62,3 @@ func (c *Client) Person(ctx Context, id int) (Person, error) {
 
 	return v, err
 }
-
-func FilterInactiveAttorneys(actors []Person) []Person {
-	var activeActors []Person
-	for _, actor := range actors {
-		if actor.SystemStatus {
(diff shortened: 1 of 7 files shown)


internal/sirius/person.go
```diff
@@ -62,13 +62,3 @@ func (c *Client) Person(ctx Context, id int) (Person, error) {
...
-
-func FilterInactiveAttorneys(actors []Person) []Person {
-	var activeActors []Person
-	for _, actor := range actors {
-		if actor.SystemStatus {
-			activeActors = append(activeActors, actor)
-		}
-	}
-	return activeActors
-}
```
GemTay, line 74: I decided to move this to case.go as we're always going to want to remove all inactive attorneys on the case, and therefore we can just run the filter on the whole case rather than passing in the which persons to filter as a parameter

gregtyler (approved):

PR #301, opened 2023-04-03
VEGA-1638 remove case ref link from deputy search result

GemTay:


gregtyler (commented):
If I'm reading the ticket correctly, I think you also need to move the `/deputy-hub/...` link to where the deputy's name is (~line 71)

GemTay:
> If I'm reading the ticket correctly, I think you also need to move the `/deputy-hub/...` link to where the deputy's name is (~line 71)

I was just copying the current logic, i'm not sure about what the correct place to direct to is with deputies :/ https://github.com/ministryofjustice/opg-sirius/blob/main/front-end/lpa/src/app/views/modules/search/search-results.html#L51:~:text=%3Ca%20ng%2Dif%3D%22person.isClient,%3C/a%3E

gregtyler:
Ah yep, fair enough 👍🏽

gregtyler (approved):

PR #319, opened 2023-05-09
[VEGA-1723] Apply search filters automatically

mabely:
[VEGA-1723](https://opgtransform.atlassian.net/browse/VEGA-1723) - apply search filters without needing a manual click to reapply. 

Note re accessibility - The auto-applying search filters is a compromise on accessibility because the user would not expect a page refresh, to counter this we could have a partial page reload (the filtered results) that has aria live region, which informs the user that a portion of the page has updated, so the screen reader doesn't read out everything again (article below).
Ref: https://technology.blog.gov.uk/2014/08/14/improving-accessibility-on-gov-uk-search/

## Checklist

* [x] I have updated the end-to-end tests to work with my changes
* [ ] I have added styling for Dark Mode
* [x] I have checked that my UI changes meet accessibility standards

[VEGA-1723]: https://opgtransform.atlassian.net/browse/VEGA-1723?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web/assets/auto-apply-filter.js
+++ b/web/assets/auto-apply-filter.js
@@ -12,7 +12,7 @@ export default function autoApplyFilter() {
 
             timeout = setTimeout(() => {
                 filter.closest("form").submit();
-            }, 500)
+            }, 1000)
         })
     })
 }
--- a/web/assets/auto-apply-filter.js
+++ b/web/assets/auto-apply-filter.js
@@ -1,6 +1,8 @@
 export default function autoApplyFilter() {
     if (document.body.className.includes('js-enabled')) {
-        document.querySelector('.moj-filter form button').classList.add("govuk-!-display-none");
+        if (document.querySelector('.moj-filter')) {
+            document.querySelector('.moj-filter form button').classList.add("govuk-!-display-none");
+        }
     }
 
     const filters = document.querySelectorAll('[data-module="app-auto-apply-filter"]');
--- a/web/assets/auto-apply-filter.js
+++ b/web/assets/auto-apply-filter.js
@@ -1,0 +1,11 @@
+export default function autoApplyFilter() {
+    const filters = document.querySelectorAll("input.govuk-checkboxes__input");
+
+    filters.forEach((filter) => {
+        filter.addEventListener('click', () => {
+            setTimeout(() => {
+                document.forms["search-filters"].submit();
+            }, 500)
+        })
+    })
+}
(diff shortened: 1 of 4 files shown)


web/assets/auto-apply-filter.js
```diff
@@ -0,0 +1,11 @@
+export default function autoApplyFilter() {
+    const filters = document.querySelectorAll("input.govuk-checkboxes__input");
+
+    filters.forEach((filter) => {
+        filter.addEventListener('click', () => {
+            setTimeout(() => {
+                document.forms["search-filters"].submit();
+            }, 500)
+        })
+    })
```
gregtyler, line 10: I think we should add a buffer to the timeout, so that it will submit the form after they've not clicked anything for 500ms, rather than giving them exactly 500ms from the first click.
```suggestion
    const filters = document.querySelectorAll("input.govuk-checkboxes__input");
    let timeout = null;

    filters.forEach((filter) => {
        filter.addEventListener('click', () => {
            if (timeout !== null) clearTimeout(timeout);

            timeout = setTimeout(() => {
                document.forms["search-filters"].submit();
            }, 500)
        })
    })
```

web/template/layout/search_filter.gohtml
```diff
@@ -32,9 +32,6 @@
                 <div class="moj-filter__options">
                     <form class="form" method="get" id="search-filters">
                         <input type="hidden" name="term" value="{{ .SearchTerm }}"/>
-                        <button class="govuk-button" data-module="govuk-button" type="submit">
-                            Apply filters
-                        </button>
```
gregtyler, line 37: If we remove the button, we should do so with JavaScript so that the form is still usable if JS isn't available.

web/assets/auto-apply-filter.js
```diff
@@ -0,0 +1,11 @@
+export default function autoApplyFilter() {
+    const filters = document.querySelectorAll("input.govuk-checkboxes__input");
```
gregtyler, line 2: This should use a custom attribute for selection (e.g. `data-module="app-auto-apply-filter"`), otherwise it will be applied to any checkboxes in the application.

web/assets/auto-apply-filter.js
```diff
@@ -0,0 +1,11 @@
...
+
+    filters.forEach((filter) => {
+        filter.addEventListener('click', () => {
+            setTimeout(() => {
+                document.forms["search-filters"].submit();
```
gregtyler, line 7: You could target the filter's closest form parent, which would avoid tightly-coupling to its ID:
```suggestion
        filter.addEventListener('click', () => {
            setTimeout(() => {
                filter.closest("form").submit();
```

gregtyler (approved):

PR #323, opened 2023-05-16
Set Pact tag/branch/commit properly

gregtyler:
Move the definition of those values to new job and depend on its output in other jobs

For SP-1505 #patch

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,6 +14,40 @@ defaults:
     shell: bash
 
 jobs:
+  identify:
+    name: Identify version
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: "0"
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -207,22 +233,11 @@ jobs:
...
-      branch: ${{ steps.set-outputs.outputs.branch }}
-      tag: ${{ steps.bump_version.outputs.tag }}
+    needs: [identify, "build", "test", "lint", "acceptance-test", "cypress"]
     steps:
       - uses: actions/checkout@v3
         with:
           fetch-depth: "0"
-      - name: Extract branch name
-        run: |
-          if [ "${{ github.head_ref }}" == "" ]; then
```
sixdaysandy, line 220: Can we keep this format for identitying the branch name as it's compatible with workflow dispatch and the other isn't

sixdaysandy (approved):

PR #334, opened 2023-06-07
VEGA-1774 add warnings to cases

GemTay:
VEGA-1774 #minor

## Checklist

* [x] I have updated the end-to-end tests to work with my changes
* [ ] I have added styling for Dark Mode - N/A
* [x] I have checked that my UI changes meet accessibility standards


internal/server/warning.go
```diff
@@ -36,17 +38,33 @@ func Warning(client WarningClient, tmpl template.Template) Handler {
 			return err
 		}
 
+		donor, err := client.Person(ctx, personId)
```
townxelliot, line 41: Does this mean a donor is now always required to be able to create a warning? Are there any situations where we create a warning without a donor? (I guess a warning is always on a case, and a case always has a donor, though?) (Just for my understanding.)

web/template/warning.gohtml
```diff
@@ -17,6 +17,22 @@
       <form class="form" method="POST">
         <input type="hidden" name="xsrfToken" value="{{ .XSRFToken }}"/>
 
+        {{ if gt (len .Donor.Cases) 1 }}
+          <div class="govuk-form-group">
+            <fieldset class="govuk-fieldset">
+              <legend class="govuk-fieldset__legend">What case is the warning for?</legend>
```
townxelliot, line 23: I'd probably use "Which case" rather than "What case" here.

townxelliot (approved):

internal/server/warning.go
GemTay: Up until now, warnings have always been created on a donor (by adding the url param ID to the POST request, except now i'm just fetching the donor for extra case data here). Yeah you've got it, a donor can exist without a case, but a case can't exist without a donor

gregtyler (dismissed):

gregtyler (approved):

PR #336, opened 2023-06-09
Use dropdown for country selector

gregtyler:
Rather than country fields being a free text field, use a dropdown populated from the "country" ref data.

I also removed some duplication in the refdata tests: they're all test cases of the same basic operation so can be done in one test harness.

Fixes VEGA-1876 #minor

### DIFF
--- a/docker/pact-stub/pact_stub.go
+++ b/docker/pact-stub/pact_stub.go
@@ -216,11 +216,9 @@ func getIgnoredInteractionsForCypressTesting(interactions []Interaction) []Inter
 	// We need this test to produce specific data in the response so that Cypress tests will pass.
 	// Since Pact won't let us return multiple array entries from `dsl.EachLike`
 	// we modify pact stub to return the desired output.
-	var updatedInteractions []Interaction
-	for _, i := range interactions {
-		switch i.ProviderState {
-		case "Some document template ids types exist":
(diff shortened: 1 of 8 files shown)


docker/pact-stub/pact_stub.go
```diff
@@ -216,19 +216,17 @@ func getIgnoredInteractionsForCypressTesting(interactions []Interaction) []Inter
...
 	// Since Pact won't let us return multiple array entries from `dsl.EachLike`
 	// we modify pact stub to return the desired output.
-	var updatedInteractions []Interaction
-	for _, i := range interactions {
-		switch i.ProviderState {
-		case "Some document template ids types exist":
-			i.Response.Body = []map[string]interface{}{{
+	for index, interaction := range interactions {
+		if interaction.Request.Path == "/lpa-api/v1/reference-data/documentTemplateId" {
+			interactions[index].Response.Body = []map[string]interface{}{{
```
gregtyler, line 221: Happened across this bit of code and wanted to tidy it up a bit so it's not dependent on provider state text

internal/sirius/ref_data_test.go
gregtyler, line 1: I've merged all the test cases into the same harness so that we're not repeating ourselves with every single test

townxelliot (approved):

PR #338, opened 2023-06-13
Use wiremock to mock requests

gregtyler:


docker/sirius-lpa-frontend/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:20.3.0-alpine3.17 as asset-env
+FROM node:20.2.0-alpine3.17 as asset-env
```
townxelliot, line 1: Why is this downgraded?

cypress/support/commands.js
```diff
@@ -1,25 +1,22 @@
...
-// Cypress.Commands.add('drag', { prevSubject: 'element'}, (subject, options) => { ... })
-//
-//
-// -- This is a dual command --
-// Cypress.Commands.add('dismiss', { prevSubject: 'optional'}, (subject, options) => { ... })
-//
-//
-// -- This will overwrite an existing command --
-// Cypress.Commands.overwrite('visit', (originalFn, url, options) => { ... })
+Cypress.Commands.add("addMock", async (url, method, response) => {
```
townxelliot, line 1: This is really nice and much easier than Mountebank. It looks especially suited to "fire and forget" scenarios where we want to mock out a single response. How does it cope if you need to provide different responses in order, e.g. a 403 for this response on the first request, then a 200 for the next one? Is there an implicit FIFO queue or something?

cypress/mocks/sirius.json
```diff
@@ -0,0 +1 @@
...tion","request":{"method":"DELETE","url":"/lpa-api/v1/payments/124"},"response":{"status":204,"headers":{}}},{"name":"A request to create a task","request":{"method":"POST","url":"/lpa-api/v1/cases/800/tasks"},"response":{"status":201,"headers":{"Content-Type":"application/json"}}},{"name":"A request to create a task for a team","request":{"method":"POST","url":"/lpa-api/v1/cases/800/tasks"},"response":{"status":201,"headers":{"Content-Type":"application/json"}}},{"name":"A request to create a person reference","request":{"method":"POST","url":"/lpa-api/v1/persons/189/references"},"response":{"status":201,"headers":{"Content-Type":"application/json"}}},{"name":"A request to create a warning","request":{"method":"POST","url":"/lpa-api/v1/warnings"},"response":{"status":201,"headers":{"Content-Type":"application/json"}}},{"name":"A search for the deleted case","request":{"method":"GET","url":"/lpa-api/v1/deleted-cases?uid=700000005555"},"response":{"status":200,"headers":{"Content-Type":"application/json"},"body":"[{\"deletedAt\":\"02/12/2022\",\"deletionReason\":\"LPA was not paid for after 12 months\",\"status\":\"Return - unpaid\",\"type\":\"LPA\",\"uId\":\"7000-0000-5555\"}]"}}]}
```
townxelliot, line 1: Is there any way to carry state between responses to mimic data stored in the session? This was something I needed to do for the govuk payments mock I wrote for Make. Also, is it possible to inject variables coming from the request back into the response? (Again, something I needed for govuk payment mocks.)

townxelliot (approved):

docker/sirius-lpa-frontend/Dockerfile
gregtyler: I was having trouble running 20.3 locally because of issues on my machine, so just downgraded to quickly unblock myself

cypress/support/commands.js
gregtyler: I don't think we can specify an order of responses, but it can do request matching and we can insert/overwrite on-the-fly between test steps, which I think should cover most scenarios.

cypress/mocks/sirius.json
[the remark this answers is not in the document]
gregtyler: This was a somewhat temporary solution to prove it worked. I think in reality I'd turn this into pretty JSON and split it across two files: one for "good" common mocks (like "/users/current" and refdata) and one for the others migrated from Pact. Then over time we can either improve the migrated ones or replace them with inline `cy.addMock`.

PR #353, opened 2023-06-29
Load a donor's cases from a specific endpoint

gregtyler:
The `/donors/:id` doesn't contain any child donor cases, making it impossible to create warnings for them.

`/persons/:id/cases` contains all of the parent and children's cases.

Fixes VEGA-1926 #patch


GemTay (approved):
🚀

PR #356, opened 2023-07-05
Use API's label property for documents and inserts

gregtyler:
Rather than using the template ID refdata.

In doing so we also remove the call to RefDataByCategory. We can also remove `DocumentTemplateRefData` because it's no longer used, and merge `DocumentTemplateTypes` into `DocumentTemplates` because it's just a sorted version of that.

We can also remove the `onScreenSummary` property, including taking it out of the API contract.

#patch


GemTay (approved):
Nice clean up! 😸

PR #365, opened 2023-08-09
SP-1663 - Switch to docker compose V2 #minor

sixdaysandy:


.prettierignore
```diff
@@ -1 +1,2 @@
 web/static
+.github/**
```
gregtyler, line 2: Was Prettier causing a problem? Or just a pain?

gregtyler (dismissed):

gregtyler (approved):

PR #376, opened 2023-08-30
VEGA-1990 MLPA documents tab

GemTay:
VEGA-1990 #minor

## Checklist

~~- [ ] I have updated the end-to-end tests to work with my changes~~
~~- [ ] I have added styling for Dark Mode~~ No changes required

### DIFF
--- a/internal/templatefn/fn.go
+++ b/internal/templatefn/fn.go
@@ -57,7 +57,7 @@ func All(siriusPublicURL, prefix, staticHash string) map[string]interface{} {
 
 			return time.Format(format), nil
 		},
-		"parseDateWithTime": func(s string, format string) (string, error) {
+		"formatDateWithTime": func(s string, format string) (string, error) {
 			t, err := time.Parse("02/01/2006 15:04:05", s)
 			if err != nil {
 				return "", err
--- a/internal/templatefn/fn.go
+++ b/internal/templatefn/fn.go
@@ -57,6 +57,14 @@ func All(siriusPublicURL, prefix, staticHash string) map[string]interface{} {
 
 			return time.Format(format), nil
 		},
+		"parseDateWithTime": func(s string, format string) (string, error) {
+			t, err := time.Parse("02/01/2006 15:04:05", s)
+			if err != nil {
+				return "", err
+			}
+
+			return t.Format(format), nil
+		},
 		"translateRefData": func(types []sirius.RefDataItem, tmplHandle string) string {
 			for _, refDataType := range types {
 				if refDataType.Handle == tmplHandle {
--- a/web/template/mlpa-documents.gohtml
+++ b/web/template/mlpa-documents.gohtml
@@ -8,9 +8,20 @@
 
             {{ template "mlpa-header" (caseTabs .Lpa "documents") }}
 
-            <div class="moj-button-menu__wrapper">
-                <h1 class="govuk-heading-l">Documents</h1>
-                <button class="govuk-button govuk-button--secondary" data-module="govuk-button">Create a document</button>
+            <div class="moj-page-header-actions">
+                <div class="moj-page-header-actions__title">
+                    <h1 class="govuk-heading-l">Documents</h1>
+                </div>
+
+                <div class="moj-page-header-actions__actions">
+                    <div class="moj-button-menu">
+                        <div class="moj-button-menu__wrapper">
+                            <button class="govuk-button moj-button-menu__item govuk-button--secondary moj-page-header-actions__action" data-module="govuk-button">
+                                Create a document
+                            </button>
+                        </div>
+                    </div>
+                </div>
             </div>
 
             <div class="govuk-body">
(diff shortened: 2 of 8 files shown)


web/template/mlpa-documents.gohtml
```diff
@@ -0,0 +1,64 @@
...
+                            {{ else if  eq .Direction "Outgoing" }}
+                                [OUT]
+                            {{ else }}
+                                [INT]
+                            {{ end }}
+                            {{ .Correspondent.PersonType }}, {{ .Correspondent.Firstname }} {{.Correspondent.Surname }}
+                            <br><a class="govuk-link" href="#">{{ .FriendlyDescription }}</a>
+                        </td>
+                        <td class="govuk-table__cell">{{ .SystemType }}</td>
+                        <td class="govuk-table__cell">{{ .CreatedDate }}</td>
```
gregtyler, line 56: I was going to point out that the design doesn't include the time but also the design doesn't match [the style guide](https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#dates), so I think it should probably be this:
```suggestion
                        <td class="govuk-table__cell">{{ (date .CreatedDate "2 January 2006") }}</td>
```

web/template/mlpa-documents.gohtml
```diff
@@ -0,0 +1,64 @@
...
+            <div class="govuk-!-display-block">
+                <h1 class="govuk-heading-l govuk-!-margin-bottom-5 govuk-!-display-inline govuk-!-padding-right-4">Documents</h1>
+                <button class="govuk-button govuk-button--secondary" data-module="govuk-button">Create a document</button>
+            </div>
+
+            <a class="govuk-link govuk-!-padding-right-4" href="#">
+                Download
+                <span class="govuk-visually-hidden">Download selected documents</span>
+            </a>
+            <a class="govuk-link app-!-colour-text-red" href="#">Delete</a>
```
gregtyler, line 20: These should be wrapped inside a `govuk-body` div, otherwise their font size is unpredictable. Also the visually hidden text is read out with the context, so need to avoid it saying "Download" twice:
```suggestion
            <div class="govuk-body">
                <a class="govuk-link govuk-!-padding-right-4" href="#">
                    Download
                    <span class="govuk-visually-hidden">selected documents</span>
                </a>
                <a class="govuk-link app-!-colour-text-red" href="#">
                    Delete
                    <span class="govuk-visually-hidden">selected documents</span>
                </a>
            </div>
```

web/template/mlpa-documents.gohtml
```diff
@@ -0,0 +1,64 @@
...
+{{ define "main" }}
+    <div class="govuk-grid-row" id="documents">
+        <div class="govuk-grid-column-two-thirds">
+
+            {{ template "mlpa-header" (caseTabs .Lpa "documents") }}
+
+            <div class="govuk-!-display-block">
+                <h1 class="govuk-heading-l govuk-!-margin-bottom-5 govuk-!-display-inline govuk-!-padding-right-4">Documents</h1>
+                <button class="govuk-button govuk-button--secondary" data-module="govuk-button">Create a document</button>
+            </div>
```
gregtyler, line 14: There's an [MOJ component for this](https://design-patterns.service.justice.gov.uk/components/page-header-actions/) that might be safer than managing the padding manually

web/template/mlpa-documents.gohtml
```diff
@@ -0,0 +1,69 @@
...
+                    <th class="govuk-table__header" scope="col">Created</th>
+                </tr>
+                </thead>
+                <tbody class="govuk-table__body">
+                {{ range .Documents }}
+                    <tr class="govuk-table__row govuk-table__row--selected">
+                        <td class="govuk-table__cell">
+                            <div class="govuk-checkboxes__item govuk-checkboxes--small moj-multi-select__checkbox">
+                                <input type="checkbox" class="govuk-checkboxes__input" id="mountain-aconcagua">
+                                <label class="govuk-label govuk-checkboxes__label" for="mountain-aconcagua">
```
gregtyler, line 44: These IDs need to vary by row so they have 1-to-1 matches, e.g.
```suggestion
                                <input type="checkbox" class="govuk-checkboxes__input" id="select-{{ .UUID }}">
                                <label class="govuk-label govuk-checkboxes__label" for="select-{{ .UUID }}">
```

internal/templatefn/fn.go
```diff
@@ -57,6 +57,14 @@ func All(siriusPublicURL, prefix, staticHash string) map[string]interface{} {
 
 			return time.Format(format), nil
 		},
+		"parseDateWithTime": func(s string, format string) (string, error) {
```
gregtyler, line 60: I think "formatDateWithTime" would be clearer.

(A can of worms here: really we should make `createdDate` a `time.Time` instance when unmarshalling the JSON, but because it returns in "Sirius date time format" we'd need to write a custom unmarshaller. The API should be using standardised date formats but we've got too much dependent code to just switch that now 😭)

web/template/mlpa-documents.gohtml
GemTay: Damn I always forget to change these once i've copied from the moj component page 😭

gregtyler (approved):

gregtyler (approved):

PR #380, opened 2023-09-06
VEGA-2017 Add basic tasks list for digital LPAs #minor

townxelliot:
The one controversial part of this is that I've added an interface to enable mocking the http client in tests. It might be that this isn't how we want to do this kind of thing.

I also struggled a bit with the unit tests, as they seem pretty clunky. I'm not sure if this is expected or I'm just doing it wrong because it's unfamiliar.

Comments/feedback on the above requested, please.

## Checklist


internal/sirius/client.go
```diff
@@ -65,12 +87,12 @@ func (c *Client) get(ctx Context, path string, v interface{}) error {
 	if err != nil {
 		return err
 	}
-	defer resp.Body.Close() //#nosec G307 false positive
 
 	if resp.StatusCode != http.StatusOK {
 		return newStatusError(resp)
 	}
 
+	defer resp.Body.Close() //#nosec G307 false positive
```
gregtyler, line 95: Was this supposed to be moved? If there's a non-200 error code then I think this now means the body won't get closed

internal/sirius/task_test.go
```diff
@@ -82,6 +95,142 @@ func TestTask(t *testing.T) {
...
+	assert.Equal(t, "Networking issue", err.Error())
+}
+
+func TestTasksForCaseBadJson(t *testing.T) {
+	mockHttpClient := &mockHttpClient{}
+
+	badJsonResponse := http.Response{
+		StatusCode: 200,
+        Body: io.NopCloser(bytes.NewBufferString("[bad json time")),
+    }
```
gregtyler, line 122: Looks like you have some mixed indentation here

cypress/mocks/migrated-from-pact.json
```diff
@@ -154,6 +154,18 @@
...
+      "request": {
+        "method": "GET",
+        "url": "/lpa-api/v1/cases/800/tasks?filter=status%3ANot+started%2Cactive%3Atrue&limit=99"
+      },
+      "response": {
+        "status": 200,
+        "headers": { "Content-Type": "application/json" },
+        "body": "{\"tasks\":[{\"id\":1,\"name\":\"Review reduced fee eligibility\",\"assignee\":{\"displayName\":\"Super Team\"}},{\"id\":2,\"name\":\"Review application correspondence\",\"assignee\":{\"displayName\":\"Marvellous Team\"}},{\"id\":3,\"name\":\"Another task\",\"assignee\":{\"displayName\":\"Super Team\"}}]}"
+      }
+    },
```
gregtyler, line 168: Since it's only used once I'd suggest creating it with `cy.addMock()` within the test

townxelliot: I didn't realise that was an option, thanks. Though I have to say I find it useful to be able to use the mock server for manual testing. Would adding it in the cypress test then make it available for manual tests?

internal/sirius/client.go
townxelliot: No, that was my mistake, I'll move it back. I probably need the same in TasksForCase().

cypress/mocks/migrated-from-pact.json
gregtyler: No the `cy.addMock` responses only work for the duration of the test they're running in. Interesting about using the mock server for testing, I prefer running it against actual Sirius so I can try things against a variety of resources.

townxelliot: I would prefer to leave this, as it works both for cypress and manual testing.

gregtyler (approved):

PR #391, opened 2023-09-20
Show incoming correspondence on Digital LPAs

gregtyler:
- Include incoming correspondence when requesting list of documents. This is done by excluding drafts and previews.
- Alter what is shown in the documents table based on document direction (e.g. showing "System Type" for outgoing docs and "type"/"subtype" for incoming)
- Add a download link to each document (previously was `#`)

## Checklist

  - N/A
  - N/A

For VEGA-2058 #minor

### DIFF
--- a/internal/sirius/document.go
+++ b/internal/sirius/document.go
@@ -3,6 +3,7 @@ package sirius
 import (
 	"fmt"
 	"sort"
+	"strings"
 )
 
 type Document struct {
@@ -14,6 +15,7 @@ type Document struct {
 	Direction           string `json:"direction"`
 	MimeType            string `json:"mimeType"`
 	SystemType          string `json:"systemType"`
+	SubType             string `json:"subType"`
 	FileName            string `json:"fileName,omitempty"`
 	Content             string `json:"content,omitempty"`
 	Correspondent       Person `json:"correspondent"`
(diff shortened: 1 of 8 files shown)


internal/sirius/document.go
```diff
@@ -27,14 +29,24 @@ const (
...
+
+	for _, docType := range docTypes {
+		query = query + "&type[]=" + docType
+	}
+
+	for _, docType := range notDocTypes {
+		query = query + "&type[-][]=" + docType
+	}
+
+	url := fmt.Sprintf("/lpa-api/v1/%s/%d/documents?%s", caseType+"s", caseId, strings.Trim(query, "&"))
```
GemTay, line 49: how come we have to specify which types of documents we want and also don't want, does just using one or the other not work correctly?

gregtyler: Outgoing documents are either Draft, Preview or Save, which is easy to manage, but incoming documents can have anything as their type.

So when we're getting a list of drafts, we want to do "is Draft". When we're displaying the list of "published" documents, it's "is not Draft nor Preview".

Really I think we shouldn't be mixing incoming and outgoing like this, but we're obviously way past that decision now 😅

GemTay (approved):

PR #393, opened 2023-09-25
Record `http.target` on trace requests

gregtyler:
Otherwise it looks like every request is to `/`

Fixes VEGA-2077 #patch


gregtyler:
Before:

image

After:

image

GemTay (approved):

PR #395, opened 2023-09-29
VEGA-2007 display linked digital lpas

GemTay:
added linked cases tabs to digital lpa case page. The cases are sorted so that their order remains the same when switching between case tabs. 

VEGA-2007 #patch

## Checklist

- [ ] I have updated the end-to-end tests to work with my changes - N/A

### DIFF
--- a/internal/templatefn/fn.go
+++ b/internal/templatefn/fn.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"html/template"
 	"net/url"
+	"sort"
 	"strings"
 	"time"
 
@@ -97,7 +98,7 @@ func All(siriusPublicURL, prefix, staticHash string) map[string]interface{} {
 				return "turquoise"
 			case "pending":
 				return "blue"
-			case "payment pending", "reduced fees pending":
+			case "payment pending", "reduced fees pending", "draft":
 				return "purple"
 			case "cancelled", "rejected", "revoked", "withdrawn", "return - unpaid", "deleted":
 				return "red"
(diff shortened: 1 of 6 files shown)


internal/templatefn/fn.go
```diff
@@ -134,14 +135,32 @@ func All(siriusPublicURL, prefix, staticHash string) map[string]interface{} {
...
+	var linkedCases []linkedCase
+	linkedCases = append(linkedCases, linkedCase{lpa.UID, lpa.Subtype})
+
+	for _, linkedLpa := range lpa.LinkedCases {
+		linkedCases = append(linkedCases, linkedCase{linkedLpa.UID, linkedLpa.Subtype})
+	}
+
+	sort.Slice(linkedCases, func(i, j int) bool {
+		return linkedCases[i].UID < linkedCases[j].UID
+	})
```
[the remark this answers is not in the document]
gregtyler: Makes sense, though we might want to order by something created date to avoid the order changing (as new UIDs are randomly generated, a new case could be inserted between two old ones)

web/template/layout/mlpa-header.gohtml
```diff
@@ -5,41 +5,59 @@
...
+                </li>
+                <li class="govuk-tabs__list-item {{ if eq .TabName "history" }}govuk-tabs__list-item--selected{{ end }}">
+                    <a class="govuk-tabs__tab" href="#history">
+                        History
+                    </a>
+                </li>
+            </ul>
+        </div>
+
+    </nav>
```
gregtyler, line 62: I think this should be before the tabs, otherwise it suggests to a screenreader that the tabs are also related to linked cases.

web/template/lpa.gohtml
```diff
@@ -66,11 +66,11 @@
...
 
-    <h2>Tasks</h2>
+    <h2 class="govuk-heading-m">Tasks</h2>
 
     <ul data-role="tasks-list">
       {{ range .TaskList }}
-        <li>{{ .Name }} ({{ .Assignee.DisplayName }})</li>
+        <li class="govuk-body">{{ .Name }} ({{ .Assignee.DisplayName }})</li>
       {{ end }}
     </ul>
```
gregtyler, line 75: We intentionally left this unstyled so it was obvious it wasn't "done"

PR #401, opened 2023-10-11
VEGA-2081 Add case actions dropdown

mabely:
Update the frontend with a 'case actions' dropdown with links to create a task and a warning - [VEGA-2081](https://opgtransform.atlassian.net/browse/VEGA-2081)

## Checklist

[VEGA-2081]: https://opgtransform.atlassian.net/browse/VEGA-2081?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


web/template/layout/mlpa-header.gohtml
```diff
@@ -22,41 +22,53 @@
...
+                </li>
+                <li class="govuk-tabs__list-item {{ if eq .TabName "history" }}govuk-tabs__list-item--selected{{ end }}">
+                    <a class="govuk-tabs__tab" href="#history">
+                        History
+                    </a>
+                </li>
+            </ul>
+        </div>
+
+        <div class="govuk-grid-column-one-quarter govuk-form-group">
```
gregtyler, line 64: There are two `govuk-grid-column-one-quarter govuk-form-group`s here (the div and the nav), I assume that's not intentional?

web/template/layout/mlpa-header.gohtml
```diff
@@ -22,41 +22,53 @@
...
+            <nav class="govuk-grid-column-one-quarter govuk-form-group" data-module="dropdown-menu">
```
gregtyler, line 65: We should prefix `data-module` values with "app-" to clarify their scope (similarly there are "govuk-" and "moj-" modules):
```suggestion
            <nav class="govuk-grid-column-one-quarter govuk-form-group" data-module="app-dropdown-menu">
```

web/assets/dropdown-menu.js
```diff
@@ -0,0 +1,15 @@
...
+    let dropdownMenuToggle = e.querySelector(
+      "[data-role='dropdown-menu-toggle']",
+    );
+
+    if (dropdownMenuToggle === null) {
+      return;
+    }
+
+    dropdownMenuToggle.addEventListener("change", (e) => {
+      window.location.href = encodeURI(e.target.value);
```
gregtyler, line 12: I think it would be good to add a bit of defence around the URL to avoid hijacks. It should always start with `/` but not with `//`
```suggestion
      const url = encodeURI(e.target.value);
      if (url.startsWith("//") || !url.startsWith("/")) {
        return;
      }
      
      window.location.href = url
```

gregtyler (approved):

PR #409, opened 2023-10-30
VEGA-2109 Redirect warnings and tasks

mabely:
## Checklist

### DIFF
--- a/internal/server/task.go
+++ b/internal/server/task.go
@@ -124,11 +124,11 @@ func Task(client TaskClient, tmpl template.Template) Handler {
 			} else {
 				data.Success = true
 
-				SetFlash(w, FlashNotification{
-					Title: "Task created",
-				})
-
 				if data.Case.CaseType == "DIGITAL_LPA" {
+					SetFlash(w, FlashNotification{
+						Title: "Task created",
+					})
+
 					return RedirectError(fmt.Sprintf("/lpa/%s", data.Case.UID))
 				}
 			}
--- a/internal/server/task.go
+++ b/internal/server/task.go
@@ -24,7 +24,6 @@ type taskData struct {
 	Success   bool
 	Error     sirius.ValidationError
 
-	Case             sirius.Case
 	TaskTypes        []string
 	Teams            []sirius.Team
 	Task             sirius.TaskRequest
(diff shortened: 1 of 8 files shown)


internal/server/task.go
```diff
@@ -120,6 +123,14 @@ func Task(client TaskClient, tmpl template.Template) Handler {
...
 			} else {
 				data.Success = true
+
+				SetFlash(w, FlashNotification{
+					Title: "Task created",
+				})
+
+				if data.Case.CaseType == "DIGITAL_LPA" {
+					return RedirectError(fmt.Sprintf("/lpa/%s", data.Case.UID))
+				}
```
gregtyler, line 133: We should only set the flash notification if we know we're going to show it immediately afterwards. otherwise it will show up on whatever the next digital LPA they open is, which would be confusing.
```suggestion
				if data.Case.CaseType == "DIGITAL_LPA" {
					SetFlash(w, FlashNotification{
						Title: "Task created",
					})

					return RedirectError(fmt.Sprintf("/lpa/%s", data.Case.UID))
				}
```

internal/server/warning.go
```diff
@@ -79,6 +82,15 @@ func Warning(client WarningClient, tmpl template.Template) Handler {
...
 			} else {
 				data.Success = true
+
+				SetFlash(w, FlashNotification{
+					Title: "Warning created",
+				})
+				for _, lpa := range data.Cases {
+					if lpa.CaseType == "DIGITAL_LPA" && slices.Contains(caseIDs, lpa.ID) {
+						return RedirectError(fmt.Sprintf("/lpa/%s", lpa.UID))
+					}
```
gregtyler, line 92: Same here with flash messages:
```suggestion
				for _, lpa := range data.Cases {
					if lpa.CaseType == "DIGITAL_LPA" && slices.Contains(caseIDs, lpa.ID) {
						SetFlash(w, FlashNotification{
							Title: "Warning created",
						})
						return RedirectError(fmt.Sprintf("/lpa/%s", lpa.UID))
					}
```

internal/server/warning.go
```diff
@@ -20,9 +22,10 @@ type warningData struct {
 	Success      bool
 	Error        sirius.ValidationError
 
-	WarningType string
-	WarningText string
-	Cases       []sirius.Case
+	WarningType  string
+	WarningText  string
+	Cases        []sirius.Case
+	FlashMessage FlashNotification
```
gregtyler, line 28: Don't think you need this any more
```suggestion
```

internal/server/task.go
```diff
@@ -70,6 +72,7 @@ func Task(client TaskClient, tmpl template.Template) Handler {
 				return err
 			}
 			data.Entity = caseitem.Summary()
+			data.Case = caseitem
```
gregtyler, line 75: `data` is used to pass things to the template, so I think we should avoid adding things to it that aren't used there.

Instead, you could create `var caseitem sirius.Case` above `group.Go`, replace the `:= client.Case` with `= client.Case`, and then use `caseitem` rather than `data.Case` when checking the type below.

gregtyler (approved):

PR #419, opened 2023-11-22
VEGA-1878 Give users the option to assign a task to themselves

iqpalm:
Add assign to me option when creating task VEGA-1878 #minor
## Checklist

### DIFF
--- a/internal/server/task.go
+++ b/internal/server/task.go
@@ -91,13 +91,13 @@ func Task(client TaskClient, tmpl template.Template) Handler {
 
 			switch assignTo {
 			case "me":
-			    user, err := client.GetUserDetails(ctx)
-			    if (err != nil){
-			        return err
-			    } else {
-                    task.AssigneeID = user.ID
-                    data.AssigneeUserName = user.DisplayName
-                }
+				user, err := client.GetUserDetails(ctx)
+				if (err != nil){
+					return err
+				} else {
+					task.AssigneeID = user.ID
+					data.AssigneeUserName = user.DisplayName
+				}
 			case "user":
 				parts := strings.SplitN(postFormString(r, "assigneeUser"), ":", 2)
 				if len(parts) == 2 {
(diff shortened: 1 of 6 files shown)


internal/server/task_test.go
```diff
@@ -222,6 +227,61 @@ func TestPostTask(t *testing.T) {
...
+		On("Case", mock.Anything, 123).
+		Return(sirius.Case{UID: "7000-0000-0000", CaseType: "LPA"}, nil)
+	client.
+		On("CreateTask", mock.Anything, 123, sirius.TaskRequest{
+			Type:        "Some task type",
+			DueDate:     "2022-03-04",
+			Name:        "Do this",
+			Description: "Please",
+			AssigneeID:  66,
+		})
```
gregtyler, line 250: You'll need to include a `Return(...)` method here to explain what the expected return value is. This means you're saying to Go "when you call `CreateTask` with these arguments, I expect you to return `nil`"

```suggestion
		}).
		Return(nil)
```

That's needed because the [`CreateTask` method](https://github.com/ministryofjustice/opg-sirius-lpa-frontend/blob/c60df9661f29854a6a213f87db090eab4cb6c974/internal/sirius/create_task.go#L18) has a required return type of `error`. Since this test is mocking a scenario where the request is successful, you can just make it return `nil`, which is treated as "no error occurred".

internal/server/task.go
```diff
@@ -89,6 +90,14 @@ func Task(client TaskClient, tmpl template.Template) Handler {
...
 
 			switch assignTo {
+			case "me":
+			    user, err := client.GetUserDetails(ctx)
+			    if (err != nil){
+			        return err
+			    } else {
+                    task.AssigneeID = user.ID
+                    data.AssigneeUserName = user.DisplayName
+                }
```
gregtyler, line 100: This is a minor thing, but you seem to have a mixture spaces and tabs on this lines. In `.go` files we just use tabs, so it would be good to stick with that.

This can be a pain to spot, but there should be a setting in your IDE to show whitespace, which will highlight where tabs and spaces are.

gregtyler (dismissed):

PR #423, opened 2023-11-28
VEGA-2165 add casework summary to digital LPA page

GemTay:
VEGA-2165 #minor

## Checklist

~~- [ ] I have updated the end-to-end tests to work with my changes~~ N/A

### DIFF
--- a/web/assets/main.scss
+++ b/web/assets/main.scss
@@ -38,6 +38,21 @@ $govuk-assets-path: "/lpa/frontend/assets/";
   border-bottom: none;
 }
 
+.app-\!-caseworker-summary {
+  background-color: #F2F6FC;
+  padding: 0 govuk-spacing(4) 0 govuk-spacing(4);
+  margin-bottom: govuk-spacing(2);
+}
+
+.app-\!-svg-icon {
+  margin-right: govuk-spacing(1);
+}
+
+.app-\!-table-head__no-vertical-padding th {
+  padding-bottom: 0;
+  padding-top: 0;
+}
+
 .app-\!-moj-search__button {
   background-color: $govuk-brand-colour;
   margin: 0;
--- a/web/assets/show-hide-case-summary.js
+++ b/web/assets/show-hide-case-summary.js
@@ -1,0 +1,20 @@
+export default function showHideCaseSummary(prefix) {
+    const button = document.querySelector('[data-module="app-case-summary-toggle"]',);
+    const buttonText = document.querySelector(".case-summary-toggle-text")
+    const buttonIcon = document.querySelector(".case-summary-toggle-icon")
+    const caseSummary = document.querySelector(".case-summary");
+
+    if (button && caseSummary) {
+        button.onclick = () => {
+            if (buttonText.innerText === "Hide") {
+                buttonText.innerText = "Show";
+                buttonIcon.classList.replace("govuk-accordion-nav__chevron--up", "govuk-accordion-nav__chevron--down");
+                caseSummary.hidden = true;
+            } else {
+                buttonText.innerText = "Hide";
+                buttonIcon.classList.replace("govuk-accordion-nav__chevron--down", "govuk-accordion-nav__chevron--up");
+                caseSummary.hidden = false;
+            }
+        };
+    }
+}
--- a/web/template/layout/mlpa-case-summary.gohtml
+++ b/web/template/layout/mlpa-case-summary.gohtml
@@ -1,0 +1,60 @@
+{{ define "mlpa-case-summary" }}
+    <div class="app-!-caseworker-summary">
+        <div class="govuk-grid-row">
+            <div class="govuk-grid-column-full">
+                <div class="govuk-!-display-block">
+                    <h3 class="govuk-heading-s govuk-!-display-inline govuk-!-padding-right-3">Case summary</h3>
+                    <button type="button" class="govuk-accordion__show-all govuk-!-margin-0" data-module="app-case-summary-toggle">
+                        <span class="govuk-accordion-nav__chevron govuk-accordion-nav__chevron--up case-summary-toggle-icon"></span>
+                        <span class="case-summary-toggle-text">Hide</span>
+                    </button>
+                </div>
+            </div>
+        </div>
+
+        <div class="govuk-grid-row case-summary">
+            <div class="govuk-grid-column-one-third">
+                <div class="govuk-!-display-block">
+                    <svg class="app-!-svg-icon" fill="none" xmlns="http://www.w3.org/2000/svg" viewBox="0 -2 19 19" height="20px" width="20px" overflow="visible">
+                        <circle cx="9.5" cy="9.5" r="9.5" fill="#1D70B8"/>
+                        <circle cx="9.5" fill="#fff" cy="13.8" r="1.3"/>
+                        <path fill="#fff" d="M8.4 4.5h2.2l-.2 7H8.6l-.2-7Z"/>
+                    </svg>
+                    <h3 class="govuk-heading-s govuk-!-display-inline">Warnings</h3>
+                </div>
+                <p class="govuk-body">There are none in place.</p>
+                <div class="govuk-!-display-block">
+                    <svg class="app-!-svg-icon" fill="none" xmlns="http://www.w3.org/2000/svg" viewBox="0 -2 19 19" height="20px" width="20px" overflow="visible">
+                        <circle cx="9.5" cy="9.5" r="9.5" fill="#1D70B8"/>
+                        <circle cx="9.5" fill="#fff" cy="13.8" r="1.3"/>
+                        <path fill="#fff" d="M8.4 4.5h2.2l-.2 7H8.6l-.2-7Z"/>
+                    </svg>
+                    <h3 class="govuk-heading-s govuk-!-display-inline">Complaints, objections & investigations</h3>
+                </div>
+                <p class="govuk-body">There are none in place.</p>
+            </div>
+            <div class="govuk-grid-column-two-thirds">
+                <table class="govuk-table">
+                    <thead class="govuk-table__head app-!-table-head__no-vertical-padding">
+                        <tr class="govuk-table__row">
+                            <th scope="col" class="govuk-table__header">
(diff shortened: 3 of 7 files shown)


web/assets/main.scss
```diff
@@ -38,6 +38,21 @@ $govuk-assets-path: "/lpa/frontend/assets/";
...
 }
 
+.app-\!-caseworker-summary {
+  background-color: #F2F6FC;
+  padding: 0 govuk-spacing(4) 0 govuk-spacing(4);
+  margin-bottom: govuk-spacing(2);
+}
+
+.app-\!-svg-icon {
+  margin-right: govuk-spacing(1);
```
GemTay, line 48: For spacing between the icon and the inline text eg "Warnings"

web/template/lpa.gohtml
```diff
@@ -3,164 +3,164 @@
...
-          {{ if not (eq .Lpa.Application.DonorAddress.Line2 "") }}<br>{{ .Lpa.Application.DonorAddress.Line2 }}{{ end }}
-          {{ if not (eq .Lpa.Application.DonorAddress.Line3 "") }}<br>{{ .Lpa.Application.DonorAddress.Line3 }}{{ end }}
-        </dd>
-      </div>
-      <div class="govuk-summary-list__row">
-        <dt class="govuk-summary-list__key">Town or city</dt>
-        <dd class="govuk-summary-list__value">{{ .Lpa.Application.DonorAddress.Town }}</dd>
-      </div>
-      {{ if not (eq .Lpa.Application.DonorAddress.Postcode "") }}
+      <dl class="govuk-summary-list">
```
GemTay, line 83: I havent changed any of this HTML, I only moved the two outer tags `<div class="govuk-grid-row">
  <div class="govuk-grid-column-full">` originally on line 6 & 7 down to line 17 & 18

web/assets/main.scss
```diff
@@ -38,6 +38,21 @@ $govuk-assets-path: "/lpa/frontend/assets/";
...
+  background-color: #f2f6fc;
+  padding: 2px govuk-spacing(4) 4px govuk-spacing(4); // the 2px top 4px bottom is so that on click/hover of the show/hide button, the box shadow doesnt exceed the height of the case summary when collapsed
+  margin: 0 govuk-spacing(-3) govuk-spacing(2) govuk-spacing(-3);
+}
+
+.app-\!-svg-icon {
+  margin-right: govuk-spacing(1);
+}
+
+.app-\!-table-head__no-vertical-padding th {
```
gregtyler, line 51: This is really picky, but if you're doing BEM then you should use `--` for modifiers and `__` for children. e.g. `.button__icon--selected`

```suggestion
.app-\!-table-head--no-vertical-padding th {
```

web/template/layout/mlpa-case-summary.gohtml
```diff
@@ -0,0 +1,60 @@
+{{ define "mlpa-case-summary" }}
+    <div class="app-!-caseworker-summary">
+        <div class="govuk-grid-row">
+            <div class="govuk-grid-column-full">
+                <div class="govuk-!-display-block">
+                    <h3 class="govuk-heading-s govuk-!-display-inline govuk-!-padding-right-3">Case summary</h3>
```
gregtyler, line 6: I think this should be a `h2` so it has the hierarchy [Donor name] -> Case summary -> Warnings
```suggestion
                    <h2 class="govuk-heading-s govuk-!-display-inline govuk-!-padding-right-3">Case summary</h2>
```

web/template/layout/mlpa-case-summary.gohtml
```diff
@@ -0,0 +1,60 @@
...
+                        <span class="case-summary-toggle-text">Hide</span>
+                    </button>
+                </div>
+            </div>
+        </div>
+
+        <div class="govuk-grid-row case-summary">
+            <div class="govuk-grid-column-one-third">
+                <div class="govuk-!-display-block">
+                    <svg class="app-!-svg-icon" fill="none" xmlns="http://www.w3.org/2000/svg" viewBox="0 -2 19 19" height="20px" width="20px" overflow="visible">
```
gregtyler, line 18: Rather than putting in a specific height:
```suggestion
                    <svg class="app-!-svg-icon" fill="none" xmlns="http://www.w3.org/2000/svg" viewBox="0 -2 19 19" overflow="visible">
```
I think it should just be matched to the height of the text surrounding it:
```css
.app-svg-icon {
  height: 1em;
}
```
That gives more flexibility for different type sizes, and makes sure it works if the user increases/reduces the font size.

web/assets/main.scss
```diff
@@ -38,6 +38,21 @@ $govuk-assets-path: "/lpa/frontend/assets/";
   border-bottom: none;
 }
 
+.app-\!-caseworker-summary {
```
gregtyler, line 41: These new classes don't need `!-` prefix: that should only be used for override classes (the rest of this file has set a bad example because it's mostly wrong too)
```suggestion
.app-caseworker-summary {
```

web/template/layout/mlpa-case-summary.gohtml
```diff
@@ -0,0 +1,60 @@
...
+                                </svg>
+                                Tasks
+                            </th>
+                            <th scope="col" class="govuk-table__header">Due date</th>
+                            <th scope="col" class="govuk-table__header">Actions</th>
+                        </tr>
+                    </thead>
+                    <tbody class="govuk-table__body">
+                        <tr class="govuk-table__row">
+                            <td class="govuk-table__cell app-!-table-row__no-border">There are no live tasks.</td>
```
gregtyler, line 53: You can make the cell span across three columns to ensure it doesn't wrap too early
```suggestion
                            <td class="govuk-table__cell app-!-table-row__no-border" colspan="3">There are no live tasks.</td>
```

web/assets/show-hide-case-summary.js
```diff
@@ -0,0 +1,28 @@
+export default function showHideCaseSummary(prefix) {
+  const button = document.querySelector(
+    '[data-module="app-case-summary-toggle"]',
+  );
+  const buttonText = document.querySelector(".case-summary-toggle-text");
+  const buttonIcon = document.querySelector(".case-summary-toggle-icon");
+  const caseSummary = document.querySelector(".case-summary");
+
+  if (button && buttonText && buttonIcon && caseSummary) {
+    button.onclick = () => {
```
gregtyler, line 10: Better to use `addEventListener` to allow each button to have multiple events:
```suggestion
    button.addEventListener('click', () => {
```

web/template/layout/mlpa-case-summary.gohtml
```diff
@@ -0,0 +1,60 @@
...
+                        <span class="govuk-accordion-nav__chevron govuk-accordion-nav__chevron--up case-summary-toggle-icon"></span>
+                        <span class="case-summary-toggle-text">Hide</span>
+                    </button>
+                </div>
+            </div>
+        </div>
+
+        <div class="govuk-grid-row case-summary">
+            <div class="govuk-grid-column-one-third">
+                <div class="govuk-!-display-block">
```
gregtyler, line 17: `div` are already block elements so shouldn't need this class
```suggestion
                <div>
```

web/assets/show-hide-case-summary.js
```diff
@@ -0,0 +1,28 @@
+export default function showHideCaseSummary(prefix) {
+  const button = document.querySelector(
+    '[data-module="app-case-summary-toggle"]',
+  );
+  const buttonText = document.querySelector(".case-summary-toggle-text");
```
gregtyler, line 5: Rather than querying `document`, it would be better to get the container for the whole case summary and look inside that. This reduces the chance that you pick up an unrelated element with a similar class.
```js
const container = button.closest('.app-caseworker-summary');
const buttonText = container.querySelector(".case-summary-toggle-text");
```

It would also be better to use custom attributes for selectors rather than classes, partly for clarity of what they're for (e.g. I was going to suggest you could remove the "case-summary" class because it's not in the CSS) but also so the browser can optimise.

So you instead might give the button text, say, `data-id="case-summary-toggle-text"` and then use:
```js
const buttonText = container.querySelector('[data-id="case-summary-toggle-text"]');
```

web/assets/show-hide-case-summary.js
```diff
@@ -0,0 +1,28 @@
...
+
+  if (button && buttonText && buttonIcon && caseSummary) {
+    button.onclick = () => {
+      if (buttonText.innerText === "Hide") {
+        buttonText.innerText = "Show";
+        buttonIcon.classList.replace(
+          "govuk-accordion-nav__chevron--up",
+          "govuk-accordion-nav__chevron--down",
+        );
+        caseSummary.hidden = true;
```
gregtyler, line 17: Should also set ariaExpanded for screenreaders:
```suggestion
        caseSummary.hidden = true;
        caseSummary.ariaExpanded = 'false';
```

web/assets/show-hide-case-summary.js
```diff
@@ -0,0 +1,28 @@
...
+      } else {
+        buttonText.innerText = "Hide";
+        buttonIcon.classList.replace(
+          "govuk-accordion-nav__chevron--down",
+          "govuk-accordion-nav__chevron--up",
+        );
+        caseSummary.hidden = false;
```
gregtyler, line 24:
```suggestion
        caseSummary.hidden = false;
        caseSummary.ariaExpanded = 'true';
```

web/template/layout/mlpa-case-summary.gohtml
```diff
@@ -0,0 +1,60 @@
+{{ define "mlpa-case-summary" }}
+    <div class="app-!-caseworker-summary">
+        <div class="govuk-grid-row">
+            <div class="govuk-grid-column-full">
+                <div class="govuk-!-display-block">
+                    <h3 class="govuk-heading-s govuk-!-display-inline govuk-!-padding-right-3">Case summary</h3>
+                    <button type="button" class="govuk-accordion__show-all govuk-!-margin-0" data-module="app-case-summary-toggle">
+                        <span class="govuk-accordion-nav__chevron govuk-accordion-nav__chevron--up case-summary-toggle-icon"></span>
+                        <span class="case-summary-toggle-text">Hide</span>
```
gregtyler, line 9: This won't show the words "case summary" on screen, but means that a screen reader will announce the button content as "Hide case summary". Otherwise if a user comes across this button they won't know what it is they're hiding.
```suggestion
                        <span class="case-summary-toggle-text">Hide</span>
                        <span class="govuk-visually-hidden">case summary</span>
```

PR #431, opened 2023-12-11
VEGA-2193 Show tasks in case summary #minor

townxelliot:
## Checklist

I've added a follow-up ticket to fix the document "save and exit" bug I discovered while working on this: https://opgtransform.atlassian.net/browse/VEGA-2207


townxelliot:
NB Code coverage appears to be incorrect, as according to local test runs, the lines added by this PR are covered by unit tests.

internal/server/create_document_digital_lpa.go
```diff
@@ -45,9 +47,17 @@ func CreateDocumentDigitalLpa(client CreateDocumentDigitalLpaClient, tmpl templa
 			XSRFToken: ctx.XSRFToken,
 		}
 
-		group, groupCtx := errgroup.WithContext(ctx.Context)
+		data.Lpa, err = client.DigitalLpa(ctx, uid)
```
gregtyler, line 50: Might be worth creating a shared function that returns `CaseTabData` so you don't have to fetch/compile the bits in each endpoint handler (or mock them each in tests).

gregtyler: Yeah, something like that. Probably easier as a follow-up ticket to give some space to explore what's possible.

townxelliot: I think it will be relatively easy to parcel up the digital LPA and task list into one structure, and provide a client method to fetch that. I think I'll do that on this ticket, as I was unhappy about the amount of repetition and busy work the current approach took.

townxelliot: We could return to some more generic way to group HTTP calls in another ticket maybe. It feels like we have a lot of repetition there, with multiple HTTP calls being made in groups, sometimes multiple times in the same handler (edit document is particularly bad for this).

townxelliot:
Note codecov report is incorrect, as I added a test to cover most of case_summary.go

internal/server/get_documents.go
```diff
@@ -4,42 +4,49 @@ import (
...
 		ctx := getContext(r)
-		lpa, err := client.DigitalLpa(ctx, uid)
-		if err != nil {
-			return err
+
+		data := getDocumentsData{
+			XSRFToken: ctx.XSRFToken,
 		}
 
-		documents, err := client.Documents(ctx, "lpa", lpa.ID, []string{}, []string{sirius.TypeDraft, sirius.TypePreview})
```
GemTay, line 34: Unrelated to this ticket but I wonder why we hardcode to fetch only "lpa" documents? `caseType` can be "lpa", "epa" or "digital_lpa" - surely we want documents of all types 🤷🏼‍♀️

web/assets/main.scss
```diff
@@ -68,6 +68,10 @@ $govuk-assets-path: "/lpa/frontend/assets/";
   color: govuk-colour("red") !important;
 }
 
+.app-\!-colour-text-lighter {
```
GemTay, line 71: You don't need !- prefix as it should only be used for override classes (this file has made it confusing to follow and set a bad example because it's wrongly used in other places too). Also you'll need to add the class to dark mode and tweak the colour as its currently difficult to ready against the dark background

web/template/layout/mlpa-header.gohtml
```diff
@@ -1,23 +1,23 @@
...
+                    <a class="moj-sub-navigation__link" {{ if eq $.CaseSummary.Lpa.UID .UID }}aria-current="page"{{ end }} href="{{ prefix (printf "/lpa/%s" .UID ) }}">
+                        {{ template "status-tag" (ToLower $.CaseSummary.Lpa.Status) }} {{ ToUpper .Subtype }} {{ .UID }}
                     </a>
                 </li>
             {{ end }}
         </ul>
     </nav>
 
-    {{ template "mlpa-case-summary" }}
+    {{ template "mlpa-case-summary" $ }}
```
GemTay, line 20: is this $ just a typo?

townxelliot: No, it passes all the variables in the current scope to the child template "mlpa-case-summary". I just added it so this was explicit.

internal/server/get_documents.go
townxelliot: I don't think I changed this, but it's a good question.

townxelliot: Should we be passing the case type of the case we're displaying documents for? This does seem to work for digital LPAs after documents have been added to it, so I'm not 100% sure what this is for.

web/template/layout/mlpa-header.gohtml
townxelliot: I took it out.

web/assets/main.scss
townxelliot: I just changed this to white.

web/template/layout/mlpa-header.gohtml
townxelliot: I realised I needed to add this back in, as it doesn't work without it :)

GemTay (approved):

PR #433, opened 2023-12-12
VEGA-1853 support inline search in new header

GemTay:
Branch based on VEGA-1852 to implement header. This PR will only contain changes to support inline search once base branch is merged

## Checklist

~~- [ ] I have updated the end-to-end tests to work with my changes~~
- [ ] I have checked that my UI changes meet accessibility standards - TODO


gregtyler (approved):

docker/sirius-lpa-frontend/Dockerfile
```diff
@@ -2,6 +2,8 @@ FROM node:21.1.0-alpine3.17 as asset-env
 
 WORKDIR /app
 
+RUN apk add --no-cache git
```
townxelliot, line 5: Do we need git in this image? I think it should not be included in the production image.

web/assets/full-width-container.js
```diff
@@ -1,20 +1,15 @@
...
-      headerContainer.style.maxWidth = "none";
-      headerContainer.className =
-        headerContainer.className +
-        " govuk-!-margin-left-5 govuk-!-margin-right-5";
-      mainContainers.forEach(
-        (container) =>
-          (container.className =
-            "govuk-!-margin-left-5 govuk-!-margin-right-5"),
-      );
+    if (mainContainers) {
```
townxelliot, line 5: Can this be done in SASS/CSS rather than with JS?

package.json
```diff
@@ -20,6 +20,8 @@
     "cypress": "13.6.1",
     "govuk-frontend": "^4.4.0",
     "jquery": "^3.6.1",
+    "opg-sirius-search-ui": "https://github.com/ministryofjustice/opg-sirius-search-ui",
+    "punycode": "^2.3.1",
```
townxelliot, line 24: What do we need punycode for here? Is it a dependency of opg-sirius-search-ui?

docker/yarn/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM node:alpine3.18
+
+RUN apk add --no-cache git
```
townxelliot, line 3: Do we need git in this image?

package.json
GemTay: I believe a package in cypress depends on it, when I run `yarn why punycode` I get this output `info This module exists because "cypress#@cypress#request#tough-cookie" depends on it.`

docker/sirius-lpa-frontend/Dockerfile
GemTay: Its needed as we run yarn below in this file. I'm unsure what has caused this issue to appear now/recently but without installing git, running `make build` fails on the yarn step with the error: `error Couldn't find the binary git`

web/assets/full-width-container.js
GemTay: I need some logic to only expand the width of the header container when on the search results page, I could apply styling to the search results template which would remove the margins and make it full width, but the header would not - so its easier to then just expand everything in here

docker/sirius-lpa-frontend/Dockerfile
gregtyler: Since this is a multi-step build, git doesn't end up on the production image (which starts at `FROM alpine:3`), it's only needed when installing the NPM dependencies on the intermediate build step

gregtyler (approved):

gregtyler (approved):

PR #453, opened 2024-01-11
VEGA-2241 display LPA store data on LPA details page

GemTay:
VEGA-2241 #patch

### DIFF
--- a/cypress/e2e/digital-lpa.cy.js
+++ b/cypress/e2e/digital-lpa.cy.js
@@ -1,5 +1,54 @@
 describe("View a digital LPA", () => {
   beforeEach(() => {
+    cy.addMock("/lpa-api/v1/digital-lpas/M-DIGI-LPA3-3333", "GET", {
+      status: 200,
+      body: {
+        uId: "M-DIGI-LPA3-3333",
+        "opg.poas.sirius": {
+          id: 333,
+          uId: "M-DIGI-LPA3-3333",
+          status: "Draft",
+          caseSubtype: "pfa",
+          createdDate: "31/10/2023",
+          investigationCount: 2,
+          complaintCount: 1,
+          taskCount: 2,
+          warningCount: 4,
+          donor: {
+            id: 33,
+          },
+          application: {
+            donorFirstNames: "Agnes",
+            donorLastName: "Hartley",
+            donorDob: "27/05/1998",
+            donorEmail: "agnes@host.example",
+            donorPhone: "073656249524",
+            donorAddress: {
+              addressLine1: "Apartment 3",
+              addressLine2: "Gherkin Building",
+              addressLine3: "33 London Road",
+              country: "GB",
+              postcode: "B15 3AA",
+              town: "Birmingham",
+            },
+            correspondentFirstNames: "Kendrick",
+            correspondentLastName: "Lamar",
+            correspondentAddress: {
+              addressLine1: "Flat 3",
+              addressLine2: "Digital LPA Lane",
+              addressLine3: "Somewhere",
+              country: "GB",
+              postcode: "SW1 1AA",
+              town: "London",
+            },
+          },
+        },
+        "opg.poas.lpastore": {
+          lpaType: "pf",
+          registrationDate: "2022-12-18",
+        },
+      },
+    });
     cy.visit("/lpa/M-DIGI-LPA3-3333");
   });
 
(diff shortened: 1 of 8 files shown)


cypress/e2e/digital-lpa.cy.js
```diff
@@ -1,5 +1,54 @@
...
+              town: "London",
+            },
+          },
+        },
+        "opg.poas.lpastore": {
+          lpaType: "pf",
+          registrationDate: "2022-12-18",
+        },
+      },
+    });
```
GemTay, line 51: Moved from `cypress/mocks/migrated-from-pact.json` as agreed in dev catchup to make it easier to see which mock responses are used with each test

internal/server/get_lpa_details.go
```diff
@@ -0,0 +1,44 @@
...
+		ctx := getContext(r)
+
+		caseSummary, err := client.CaseSummary(ctx, uid)
+
+		if err != nil {
+			return err
+		}
+
+		var lpaStoreData map[string]interface{}
+		_ = json.Unmarshal(caseSummary.DigitalLpa.LpaStoreData, &lpaStoreData)
```
GemTay, line 35: `json.RawMessage` returns the data as a byte array, so I have to unmarshal it make it readable

GemTay:
image

cypress/e2e/digital-lpa.cy.js
```diff
@@ -1,5 +1,54 @@
 describe("View a digital LPA", () => {
   beforeEach(() => {
```
townxelliot, line 2: Does the mock need to be added before each test, or just before the suite? I think a `before()` property will only be called once for the suite.

cypress/mocks/migrated-from-pact.json
```diff
@@ -97,18 +97,6 @@
         "body": "{\"opg.poas.sirius\":{\"application\":{\"correspondentAddress\":{\"addressLine1\":\"Main Line Bungalow\",\"addressLine2\":\"Himmerton Lane\",\"addressLine3\":\"Sutton\",\"country\":\"GB\",\"postcode\":\"S24 7DJ\",\"town\":\"Scarsdale\"},\"correspondentFirstNames\":\"Heath\",\"correspondentLastName\":\"Enstad\",\"donorAddress\":{\"addressLine1\":\"Apartment 24\",\"addressLine2\":\"Navigation Building\",\"addressLine3\":\"90 London Road\",\"country\":\"GB\",\"postcode\":\"B15 4TA\",\"town\":\"Birmingham\"},\"donorDob\":\"27/05/1978\",\"donorEmail\":\"zswanberg@host.example\",\"donorFirstNames\":\"Zoraida\",\"donorLastName\":\"Swanberg\",\"donorPhone\":\"073456249524\"},\"caseSubtype\":\"pfa\",\"complaintCount\":1,\"createdDate\":\"31/08/2023\",\"investigationCount\":2,\"status\":\"Draft\",\"taskCount\":3,\"id\":900,\"warningCount\":4},\"uId\":\"M-9999-4567-AAAA\"}"
       }
     },
-    {
```
townxelliot, line 100: Is this definitely not used by any of the other cypress tests?

internal/server/get_lpa_details.go
townxelliot, line 35: If you're not doing anything with the return value, I think it can be omitted altogether. Though I'm wondering if we should check whether there is an error here, just in case the JSON from the store is invalid.

townxelliot (approved):

cypress/e2e/digital-lpa.cy.js
GemTay: Looks like we [reset the mocks](https://github.com/ministryofjustice/opg-sirius-lpa-frontend/blob/main/cypress/support/index.js) after each test so `before()` makes only the first test work and others fail. I'm not entirely sure why we reset the mocks after each test, but i'm guessing its so that each test environment is refreshed so that tests are independent and isolated, and not dependant on each other to avoid a domino affect

cypress/mocks/migrated-from-pact.json
GemTay: I'm fairly certain it isn't, only in the digital lpa cypress tests do we perform the API call `/lpa-api/v1/digital-lpas/M-DIGI-LPA3-3333`

internal/server/get_lpa_details.go
GemTay: good shout! I'll add the error handling in

townxelliot (approved):

PR #479, opened 2024-02-22
Switch logger to slog, use new logging standards

gregtyler:
For CTC-138 #minor


main.go
```diff
@@ -2,84 +2,51 @@ package main
...
-	otel.SetTextMapPropagator(xray.Propagator{})
-
-	return func() {
-		if err := tp.Shutdown(ctx); err != nil {
-			logger.Fatal(err)
-		}
-	}
+func fatal(err error, logger *slog.Logger) {
+	logger.Error(fmt.Sprintf("a fatal error occurred: %s", err.Error()))
+	os.Exit(1)
```
hawx, line 25: You may find it nicer to structure with `main` calling something else like https://github.com/ministryofjustice/opg-modernising-lpa/blob/main/cmd/mlpa/main.go#L73, that way `defer`s will happen as expected

main.go
```diff
@@ -2,84 +2,51 @@ package main
...
-	otel.SetTracerProvider(tp)
-	otel.SetTextMapPropagator(xray.Propagator{})
-
-	return func() {
-		if err := tp.Shutdown(ctx); err != nil {
-			logger.Fatal(err)
-		}
-	}
+func fatal(err error, logger *slog.Logger) {
+	logger.Error(fmt.Sprintf("a fatal error occurred: %s", err.Error()))
```
hawx, line 24: I wrote ours in the style of `logger.Error("a fatal error occurred", slog.Any("err", err))`, not sure if you'd prefer that

hawx (dismissed):

main.go
gregtyler: Yours is more structured, potentially easier to filter for, I'll switch over 👍🏽

MishNajam (approved):

PR #505, opened 2024-04-18
vega-2443 Prevent case type cell from wrapping

mabely:
[vega-2443](https://opgtransform.atlassian.net/browse/VEGA-2443)

## Checklist

### DIFF
--- a/web/template/search.gohtml
+++ b/web/template/search.gohtml
@@ -52,7 +52,7 @@
                         <tbody class="govuk-table__body">
                         {{ range $r := (slice .Results 0 (len .Results)) }}
                             <tr class="govuk-table__row">
-                                <td class="govuk-table__cell govuk-link--no-visited-state">
+                                <td class="govuk-table__cell govuk-link--no-visited-state table-cell-nowrap">
                                     {{ range $case := .Cases }}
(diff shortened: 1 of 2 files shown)


web/template/search.gohtml
```diff
@@ -88,7 +88,7 @@
                                 </td>
                                 <td class="govuk-table__cell">{{ formatDate .DateOfBirth }}</td>
                                 <td class="govuk-table__cell">{{ .AddressSummary }}</td>
-                                <td class="govuk-table__cell">
+                                <td class="govuk-table__cell table-cell-nowrap">
                                     {{ range $case := .Cases }}
```
townxelliot, line 92: Do you have a screenshot showing what this does to the layout? I'm not sure whether we should be using linebreaks to space things out. Could we be using paragraphs instead? That should remove the need for nowrap and white space tweaks.

mattmachell: You should indeed avoid using line breaks for formatting and wrap in appropriate element instead. Might be a list in this case if it's a list of cases.

mabely: I'll make a separate ticket for that just because it passed testing earlier and it was a nightmare trying to get a dev environment up

PR #509, opened 2024-04-22
VEGA-2029 create history tab

MishNajam:
## Checklist

### DIFF
--- a/internal/server/get_history.go
+++ b/internal/server/get_history.go
@@ -11,7 +11,7 @@ import (
 )
 
 type GetHistoryClient interface {
-	GetEvents(ctx sirius.Context, id int, caseId string) (any, error)
+	GetEvents(ctx sirius.Context, id int, caseId int) (any, error)
 	CaseSummary(ctx sirius.Context, uid string) (sirius.CaseSummary, error)
 }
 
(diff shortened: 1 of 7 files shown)


internal/server/get_history.go
```diff
@@ -0,0 +1,45 @@
...
+	GetEvents(ctx sirius.Context, id int, caseId int) (any, error)
+	CaseSummary(ctx sirius.Context, uid string) (sirius.CaseSummary, error)
+}
+
+type getHistory struct {
+	CaseSummary sirius.CaseSummary
+	EventData   any
+}
+
+func GetHistory(client GetHistoryClient, tmpl template.Template) Handler {
```
gregtyler, line 21: This handler should have a test

internal/sirius/get_events.go
```diff
@@ -0,0 +1,15 @@
+package sirius
+
+import (
+	"fmt"
+)
+
+func (c *Client) GetEvents(ctx Context, id int, caseId int) (any, error) {
```
gregtyler, line 7: This should probably also have a contract test. Appreciate we're not doing much with the structure yet, so it's hard to add a contract to the body, but we could at least say "it must return an array of objects".

gregtyler (approved):

PR #523, opened 2024-05-13
VEGA-2445 Dark mode fixes for inputs, warning and inset text #minor

townxelliot:
## Checklist

- [X] I have updated the end-to-end tests to work with my changes
- [X] I have added styling for Dark Mode
- [X] I have checked that my UI changes meet accessibility standards

### DIFF
--- a/web/assets/dark.scss
+++ b/web/assets/dark.scss
@@ -91,6 +91,12 @@ $sirius-bg-color: #29292e;
     color: govuk-colour("white");
   }
 
+  .govuk-input,
+  .govuk-textarea,
+  .govuk-select {
+    border-color: govuk-colour("white");
+  }
+
   .moj-pagination__link::before,
   .moj-pagination__link::after {
     filter: invert(100%);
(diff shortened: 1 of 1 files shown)


web/assets/dark.scss
```diff
@@ -174,4 +180,11 @@ $sirius-bg-color: #29292e;
       fill: govuk-colour("white");
     }
   }
+
+  .govuk-inset-text,
+  .govuk-inset-text *,
+  .govuk-warning-text,
+  .govuk-warning-text * {
```
gregtyler, line 187: Are the `*` selectors necessary? I'm conscious it might override things we don't want it to (a contrived example is that `<div class="govuk-inset-text"><span class="govuk-hint">Hint</span></div>` is white rather than the hint text colour).

townxelliot: I could make it more specific to .gov-inset-text__text and similar, which would probably help, then make sure all places where we use the inset and warning components are correctly marked-up.

MishNajam (approved):

PR #538, opened 2024-05-30
vega-2364 - Add donor section label to progress indicators

mabely:
## Checklist


townxelliot:
The fn_test.go should be simple to extend to cover the new label translation. I'm not sure this is that valuable, but codecov will just keep complaining until we do it.

townxelliot (approved):

PR #541, opened 2024-06-06
Vega 2238 manual closing of tasks

MishNajam:
## Checklist

### DIFF
--- a/internal/server/clear_task.go
+++ b/internal/server/clear_task.go
@@ -49,6 +49,9 @@ func ClearTask(client ClearTaskClient, tmpl template.Template) Handler {
 
 		data := clearTaskData{XSRFToken: ctx.XSRFToken}
 		taskID, err := strconv.Atoi(r.FormValue("id"))
+		if err != nil {
+			return err
+		}
 
 		group.Go(func() error {
 			task, err := client.Task(ctx.With(groupCtx), taskID)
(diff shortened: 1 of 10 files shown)


internal/server/clear_task.go
```diff
@@ -0,0 +1,94 @@
...
+		ctx := getContext(r)
+		group, groupCtx := errgroup.WithContext(ctx.Context)
+
+		data := clearTaskData{XSRFToken: ctx.XSRFToken}
+		taskID, err := strconv.Atoi(r.FormValue("id"))
+		if err != nil {
+			return err
+		}
+
+		group.Go(func() error {
```
townxelliot, line 56: This doesn't need to be inside a group, as it's a single request.

Do we only need to do this request if this is a GET?

internal/server/clear_task.go
```diff
@@ -0,0 +1,94 @@
...
+			task, err := client.Task(ctx.With(groupCtx), taskID)
+			if err != nil {
+				return err
+			}
+
+			if lpa == nil && len(task.CaseItems) > 0 {
```
townxelliot, line 62: I don't think `lpa` is ever anything but nil at this point, so that condition can be removed.

internal/server/clear_task.go
```diff
@@ -0,0 +1,94 @@
...
+		if r.Method == http.MethodPost {
+			err = client.ClearTask(ctx, taskID)
+
+			if ve, ok := err.(sirius.ValidationError); ok {
+				w.WriteHeader(http.StatusBadRequest)
+				data.Error = ve
+			} else if err != nil {
+				return err
+			} else if lpa != nil && lpa.CaseType == "DIGITAL_LPA" {
+				SetFlash(w, FlashNotification{Title: "Task completed"})
```
townxelliot, line 85: Would it make sense if this said "Task cleared"? (I haven't seen the designs.)

Then again, if it's been through QA, I assume this was checked.

MishNajam: This was checked. It will be UR'd too so future changes may happen but I think it's fine for now.

townxelliot (approved):

PR #566, opened 2024-09-11
VEGA-2513 Add actors to document recipient list from LPA store

nickpyett-djg:
[VEGA-2513](https://opgtransform.atlassian.net/browse/VEGA-2513) This PR covers changing the donor details from Sirius data to LPA store data and adding other actors to the document recipient list.

## Checklist

[VEGA-2513]: https://opgtransform.atlassian.net/browse/VEGA-2513?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/server/create_document_digital_lpa.go
```diff
@@ -120,6 +182,12 @@ func CreateDocumentDigitalLpa(client CreateDocumentDigitalLpaClient, tmpl templa
 					continue
 				}
 
+				// Create contact for each recipient as we currently have no way to link a document to LPA store actors
+				recipient, err = client.CreateContact(ctx, recipient)
+				if err != nil {
+					return err
+				}
```
gregtyler, line 189: Can we wrap this in `if (recipient.ID < 0)` to avoid creating new records for recipients that already exist? I appreciate that can't happen right now, but it will be useful in the future if we ever populate `data.Recipients` with existing people records

nickpyett-djg: Good point this is done.

gregtyler (approved):

PR #649, opened 2025-04-16
VEGA-2873 : Record court order dates for severance

ndasmoj:
This PR covers [VEGA-2873](https://opgtransform.atlassian.net/browse/VEGA-2873)

I have updated the implementation of the severance flow as part of this ticket to allow the SeveranceApplication object and the bool fields `hasDonorConsented` and `severanceOrdered` to be treated as nil in Go when they are returned as empty from the API. 

In my testing I found that they are treated as empty object or false and so I changed the relevant fields in Go to pointers which then allows the object or the bool fields as nil.

## Checklist

[VEGA-2873]: https://opgtransform.atlassian.net/browse/VEGA-2873?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/sirius/edit_severance_application_test.go
+++ b/internal/sirius/edit_severance_application_test.go
@@ -51,35 +51,6 @@ func TestEditSeveranceApplication(t *testing.T) {
 					})
 			},
 		},
-		{
-			name: "Severance Ordered",
-			severanceApplication: SeveranceApplication{
-				SeveranceOrdered:       convertToBool(true),
-				CourtOrderDecisionMade: "2025-04-05",
-				CourtOrderReceived:     "2025-04-10",
-			},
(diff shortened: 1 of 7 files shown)


internal/sirius/edit_severance_application_test.go
```diff
@@ -10,22 +10,26 @@ import (
 	"testing"
 )
 
+func convertToBool(b bool) *bool {
```
gregtyler, line 13: Maybe a bit of a clearer name?
```suggestion
func boolPointer(b bool) *bool {
```

PR #666, opened 2025-05-19
VEGA-2663 : Edit Restrictions #minor

ndasmoj:
This PR covers [VEGA-2663](https://opgtransform.atlassian.net/browse/VEGA-2663) & [VEGA-2976](https://opgtransform.atlassian.net/browse/VEGA-2976)

## Checklist

- [X] I have updated the end-to-end tests to work with my changes

[VEGA-2663]: https://opgtransform.atlassian.net/browse/VEGA-2663?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[VEGA-2976]: https://opgtransform.atlassian.net/browse/VEGA-2976?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/server/manage_restrictions.go
+++ b/internal/server/manage_restrictions.go
@@ -43,7 +43,6 @@ type manageRestrictionsData struct {
 	AmendedRestrictions       string
 	FormAction                string
 	ConfirmRestrictionDetails CourtOrderRestrictionDetails
-	ExistingSeverance         *sirius.SeveranceApplication
 	Success                   bool
 }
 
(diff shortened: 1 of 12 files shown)


internal/server/manage_restrictions.go
```diff
@@ -138,41 +156,128 @@ func ManageRestrictions(client ManageRestrictionsClient, tmpl template.Template)
...
+
+					if data.CourtOrderReceivedDate != "" {
+						severanceApplication.CourtOrderReceived = data.CourtOrderReceivedDate
+					}
+
+					if data.AmendedRestrictions != "" {
+						severanceApplication.UpdatedRestrictions = data.AmendedRestrictions
+					}
+
+					switch data.SeveranceOrderedByCourt {
```
nickpyett, line 244: Does this switch statement only have one case? It might be better as an if with an or?

ndasmoj: I have kept it as a switch in case we add more cases in the future as there were discussions around it. Its the same with donor consent which is why its a switch rather than an if else.

internal/templatefn/fn.go
```diff
@@ -319,6 +319,9 @@ func All(siriusPublicURL, prefix, staticHash string) map[string]interface{} {
 				return "objection type NOT RECOGNISED: " + s
 			}
 		},
+		"compareBoolPointers": func(i *bool, j bool) bool {
```
nickpyett, line 322: This is a very small function but might be worth adding a test to `fn_test.go` as the other methods seem to have tests.

ndasmoj: Thanks. I have added the test.

PR #671, opened 2025-06-02
VEGA-2913 : Changes for adding a  case status reason

ndasmoj:
This PR covers [VEGA-2913](https://opgtransform.atlassian.net/browse/VEGA-2913)

As part of the PR, the rendering of the Case Status radios have been moved away from using the 'radio' template to allow the inclusion of the conditional radios. The conditional radios are being rendered using Ref Data response. 

## Checklist

[VEGA-2913]: https://opgtransform.atlassian.net/browse/VEGA-2913?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cypress/e2e/change-case-status.cy.js
+++ b/cypress/e2e/change-case-status.cy.js
@@ -93,11 +93,13 @@ describe("Change case status", () => {
 
     cy.addMock("/lpa-api/v1/reference-data/caseChangeReason", "GET", {
       status: 200,
-      body: [{
-        handle: "LPA_DOES_NOT_WORK",
-        label: "The LPA does not work and cannot be changed",
-        parentSources: ["cannot-register"],
-      }],
+      body: [
+        {
+          handle: "LPA_DOES_NOT_WORK",
+          label: "The LPA does not work and cannot be changed",
+          parentSources: ["cannot-register"],
+        },
+      ],
     });
 
     const mocks = Promise.allSettled([


cypress/mocks/common.json
```diff
@@ -107,6 +107,17 @@
         "body": "[{\"handle\":\"GB\",\"label\":\"Great Britain\"}]"
       }
     },
```
MishNajam, line 109: This was the old way of adding mocks, now we add our mocks into our test file. 

Something like:
```
    cy.addMock("/lpa-api/v1/reference-data/caseChangeReason", "GET", {
      status: 200,
      body: {
        handle: "LPA_DOES_NOT_WORK",
        label: "The lpa does not work...",
        parentSources: ["cannot-register"],
      },
    });
```

internal/server/change_case_status.go
```diff
@@ -41,33 +53,72 @@ func ChangeCaseStatus(client ChangeCaseStatusClient, tmpl template.Template) Han
...
-				data.Success = true
+		if err := group.Wait(); err != nil {
+			return err
+		}
+
+		if r.Method == http.MethodPost {
+			if (data.NewStatus == "cannot-register" || data.NewStatus == "cancelled") && data.StatusChangeReason == "" {
 				data.OldStatus = data.NewStatus
+				w.WriteHeader(http.StatusBadRequest)
+				data.Error.Field["changeReason"] = map[string]string{
```
MishNajam, line 95: This should be able to be handled in the backend. We have done similar logic in the create draft file. For the different addresses.

web/template/change_case_status.gohtml
```diff
@@ -19,19 +19,38 @@
...
-        (item "statutory-waiting-period" "Statutory waiting period" "statusTag" true)
-        (item "registered" "Registered" "statusTag" true)
-        (item "suspended" "Suspended" "statusTag" true)
-        (item "do-not-register" "Do not register" "statusTag" true)
-        (item "expired" "Expired" "statusTag" true)
-        (item "cannot-register" "Cannot register" "statusTag" true)
-        (item "cancelled" "Cancelled" "statusTag" true)
-        (item "de-registered" "De-registered" "statusTag" true)
-        ) }}
-
```
MishNajam, line 34: Optional: It may be cleaner to refactor the logic in the original template or create a new template for this as there will be future tickets where we have drop down radios appear.

cypress/mocks/common.json
ndasmoj: Done.

internal/server/change_case_status.go
```diff
@@ -5,30 +5,42 @@ import (
...
+	StatusChangeReason      string
 }
 
 func ChangeCaseStatus(client ChangeCaseStatusClient, tmpl template.Template) Handler {
 	return func(w http.ResponseWriter, r *http.Request) error {
 		caseUID := r.FormValue("uid")
 
 		ctx := getContext(r)
+		group, groupCtx := errgroup.WithContext(ctx.Context)
 		cs, err := client.CaseSummary(ctx, caseUID)
```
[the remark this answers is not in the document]
ndasmoj: Done

web/template/change_case_status.gohtml
ndasmoj: As discussed, I did try putting it into the template in various ways but none of them worked - they all failed at the stage of displaying different reason values based on different statuses. It needs quite a bit of work to put the conditional radios into templates, so needs to be done as a separate piece of work.

PR #684, opened 2025-06-18
VEGA-3040 : Clear task when court order instructions are received

ndasmoj:
This PR covers [VEGA-3040](https://opgtransform.atlassian.net/browse/VEGA-3040)

## Checklist

[VEGA-3040]: https://opgtransform.atlassian.net/browse/VEGA-3040?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/server/manage_restrictions.go
+++ b/internal/server/manage_restrictions.go
@@ -252,6 +252,13 @@ func ManageRestrictions(client ManageRestrictionsClient, manageTmpl template.Tem
 						return err
 					}
 
+					if taskID != 0 {
+						err := client.ClearTask(ctx, taskID)
+						if handleError(w, &data, err) {
+							return err
+						}
+					}
+
 					return handleSuccess(w, &data, caseUID)
 				}
 			}


internal/server/manage_restrictions.go
```diff
@@ -252,6 +252,13 @@ func ManageRestrictions(client ManageRestrictionsClient, manageTmpl template.Tem
 						return err
 					}
 
+					if taskID != 0 {
+						err := client.ClearTask(ctx, taskID)
+						if handleError(w, &data, err) {
+							return err
+						}
+					}
```
gregtyler, line 260: I'd suggest putting this in the API to keep the business logic together. Also because it will do it in a DB transaction so if anything fails it will rollback, whereas putting the logic here means you could end up editing the severance data but not clearing the task if the second request failed.

ndasmoj:
Closing the PR as no changes needed.

PR #727, opened 2025-08-22
Vega 3174 move to enums

Nicholassully:
## Checklist


internal/shared/case_status.go
```diff
@@ -0,0 +1,168 @@
...
+	case CaseStatusTypeDeleted:
+		return "Deleted"
+	case CaseStatusTypeRevoked:
+		return "Revoked"
+	default:
+		return ""
+	}
+}
+
+func (cs CaseStatus) CaseStatusColour() string {
```
gregtyler, line 109: Don't think we need to include "CaseStatus" in the method name because that's known from the type?
```suggestion
func (cs CaseStatus) Colour() string {
```

PR #738, opened 2025-09-05
VEGA-3175 moving case status to enum

Nicholassully:
## Checklist

### DIFF
--- a/internal/server/change_case_status.go
+++ b/internal/server/change_case_status.go
@@ -2,11 +2,14 @@ package server
 
 import (
 	"fmt"
+	"net/http"
+	"strings"
+
 	"github.com/ministryofjustice/opg-go-common/template"
+	"github.com/ministryofjustice/opg-sirius-lpa-frontend/internal/shared"
 	"github.com/ministryofjustice/opg-sirius-lpa-frontend/internal/sirius"
 	"github.com/ministryofjustice/opg-sirius-lpa-frontend/internal/templatefn"
 	"golang.org/x/sync/errgroup"
-	"net/http"
 )
 
 type ChangeCaseStatusClient interface {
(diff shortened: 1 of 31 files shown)


internal/server/change_case_status.go
```diff
@@ -70,37 +72,37 @@ func ChangeCaseStatus(client ChangeCaseStatusClient, tmpl template.Template) Han
...
 		}
 
 		data := changeCaseStatusData{
 			XSRFToken:               ctx.XSRFToken,
 			Error:                   sirius.ValidationError{Field: sirius.FieldErrors{}},
 			Entity:                  fmt.Sprintf("%s %s", cs.DigitalLpa.SiriusData.Subtype, caseUID),
 			CaseUID:                 caseUID,
-			OldStatus:               status,
-			NewStatus:               postFormString(r, "status"),
+			OldStatus:               strings.ToLower(status),
```
nickpyett, line 84: Don't think we need `ToLower` here.

internal/server/change_case_status.go
```diff
@@ -31,7 +33,7 @@ type changeCaseStatusData struct {
 	StatusItems             []statusItem
 	CaseStatusChangeReasons []sirius.RefDataItem
 	OldStatus               string
```
nickpyett, line 35: Is it worth changing the type for `OldStatus` to `shared.CaseStatus`?

internal/shared/case_status.go
```diff
@@ -0,0 +1,205 @@
...
+	CaseStatusTypeDoNotRegister
+	CaseStatusTypeExpired
+	CaseStatusTypeCannotRegister
+	CaseStatusTypeDeRegistered
+	CaseStatusTypeSuspended
+	CaseStatusTypeImperfect
+	CaseStatusTypeInvalid
+)
+
+var caseStatusTypeMap = map[string]CaseStatus{
```
nickpyett, line 34: There is also a `'With COP` status in the PowerOfAttorney\Lpa.php file in Sirius that might need to be supported.

nickpyett (approved):

PR #741, opened 2025-09-10
VEGA-3122 : add static LPA download link

davidbridgwood:
add link for static lpa.
fixes: [VEGA-3122](https://opgtransform.atlassian.net/browse/VEGA-3122)

### DIFF
--- a/web/template/mlpa-details.gohtml
+++ b/web/template/mlpa-details.gohtml
@@ -237,4 +237,5 @@
         </div>
       </div>
     {{ end }}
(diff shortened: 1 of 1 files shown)


almo2k (dismissed):
Looks good,  could the accessibility be improved by adding a aria ?

something like

```html
<a class="govuk-link" 
   id="download-static-lpa" 
   href="{{ sirius (printf "/api/v1/lpas/%s/static" .CaseSummary.DigitalLpa.UID) }}" 
   download
   aria-label="Download static LPA - original submission data file">
  Download static LPA
</a>
```

almo2k (changes requested):
Hi mate, best practice is to not use the same aria label as is in the link - the screen reader will end up reading it twice and it won't give those that need the assistance anything extra. I tend to think of it as a descriptive tag for those that are partially sighted

PR #746, opened 2025-09-23
VEGA-3222 reduce and rationalise mocks in digital-lpa cypress tests

davidbridgwood:
fixes: [VEGA-3222](https://opgtransform.atlassian.net/browse/VEGA-3222)
- Made more use of the wiremocks in `cypress/mocks/cases.js` & `digitalLpas.js`
- removed/rationalised redundant mock lpas/data
- deleted test `shows certificate provider witness only (1 witness)` as what was being tested did not match description and isn't a valid scenario
- moved mocks which were only required for a single test into the block of that test to make it clearer why its needed

[VEGA-3222]: https://opgtransform.atlassian.net/browse/VEGA-3222?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cypress/e2e/digital-lpa.cy.js
+++ b/cypress/e2e/digital-lpa.cy.js
@@ -171,8 +171,6 @@ describe("View a digital LPA", () => {
       cases.warnings.empty("1111"),
       cases.tasks.empty("1111"),
       digitalLpas.objections.empty("M-DIGI-LPA3-3333"),
-      digitalLpas.get("M-DIGI-LPA3-3335"),
-      digitalLpas.objections.empty("M-DIGI-LPA3-3335"),
     ]);
 
     cy.wrap(mocks);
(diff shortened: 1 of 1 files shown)


almo2k (commented):
Hi David,
Is the helper function not helpful? I personally think it helps with maintainability, gives us override capabilities and is easier to read. 

I'm not sure about the removal of the 1 witness scenario test - having looked at the codebase we have PHP unit tests that explicitly test testSignedOnBehalfWithOnlyOneCertificateProviderWitness(), The
Go methods WasWitnessedByCertificateProvider() and WasWitnessedByIndependentWitness() operate independently and the
Frontend template uses OR logic: {{ if or (.DigitalLpa.WasWitnessedByCertificateProvider) (.DigitalLpa.WasWitnessedByIndependentWitness) }}

If the business logic is that there can never be a single witness then we would need to look at updating these parts of the template/backend right?

PR #927, opened 2026-04-10
VEGA-3766 - Implement new create EPA form

MikeRJustice:
- Implement create EPA Go logic and template
- Add missing fields from Case struct
- Add new api call
- Add tests


internal/templatefn/fn.go
```diff
@@ -495,6 +506,18 @@ func radios(name, label string, value interface{}, errors map[string]string, ite
...
+}
+
+func inlineRadios(name, label string, value interface{}, errors map[string]string, items ...itemData) radiosData {
+	return radiosData{
+		Name:   name,
+		Label:  label,
+		Value:  value,
+		Errors: errors,
+		Items:  items,
+		Inline: true,
```
Nicholassully, line 520: could we not do this 
func radios(name, label string, value interface{}, errors map[string]string, items ...itemData, isInline bool) radiosData {
radiosData{}
  if isInline {radiosData.Inline = true}
else {radiosData.Inline = false}
  return radiosData
}

then where we want to send in a true we could just 
(radios "donorHasOtherEpas" "Has the donor made any other EPAs?" (stringifyBoolPointer .Case.DonorHasOtherEpas) true)

MikeRJustice: I believe that only the final parameter in a function can be variadic. Doing it this way removes the need to reorder the parameters in the radios function which would involve touching a lot of extra files (anywhere we use the radios function would need to be changed to include the isInline argument).

MishNajam (dismissed):

MishNajam (dismissed):

MishNajam (approved):

MishNajam (approved):

PR #932, opened 2026-04-15
VEGA-3773 - Implement new edit epa form

MikeRJustice:
Fixes [VEGA-3773](https://opgtransform.atlassian.net/browse/VEGA-3773)

- Implement the edit EPA form
- Refactor the EPA struct to include only EPA specific fields with an embedded Case struct
- Add tests

[VEGA-3773]: https://opgtransform.atlassian.net/browse/VEGA-3773?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/server/create_epa.go
+++ b/internal/server/create_epa.go
@@ -85,8 +85,9 @@ func CreateEpa(client CreateEpaClient, tmpl template.Template) Handler {
 			data.Epa = epa
 			data.AppointmentType = caseAttorneyValue
 
+			var caseId int
 			if isEditing {
-				caseId, err := strToIntOrStatusError(caseIdStr)
+				caseId, err = strToIntOrStatusError(caseIdStr)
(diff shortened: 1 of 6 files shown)


MishNajam:
Mainly LGTM but would be good if we could bump up some of the test coverage, a test to just check an error occurs if the Epa endpoint fails could be a quick win as I think there should be examples on the codebase. Also you can use [table driven testing](https://github.com/ministryofjustice/opg-sirius-lpa-frontend/blob/455b1bf868eb104f9e87d8d828d06c5f548c00b3/internal/server/document_list_test.go#L176) to repeat your test with different appointTypes

MishNajam (approved):

PR #954, opened 2026-05-06
vega-3854 adding missing tags to dark mode

Nicholassully:
## Checklist


web/assets/dark.scss
davidbridgwood, line 1: The ticket mentions a couple of other missing classes

-  `govuk-accordion__section-toggle` which i think should be set in the same block as `govuk-accordion__show-all` so they are consistent in the accordion.

- `.govuk-button` but im not sure what the styling should be for that one.

davidbridgwood (approved):

PR #960, opened 2026-05-11
VEGA-3869 - Copy to clipboard button POC

MikeRJustice:
Fixes [VEGA-3869](https://opgtransform.atlassian.net/browse/VEGA-3869)

- Adds a javascript function for copying text to the clipboard. The function copies the given string, adds a hidden screen reader alert, and changes the copy button to a non-interactive element that reads `Copied`. After 4 seconds, the copy button reverts back to it's original state.
- Adds the copy button to the case id in the timeline event just as an example. It can be used anywhere.
  

[VEGA-3869]: https://opgtransform.atlassian.net/browse/VEGA-3869?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web/assets/copy-to-clipboard.js
+++ b/web/assets/copy-to-clipboard.js
@@ -27,5 +27,5 @@ export default async function copyToClipboard() {
 
       copyButton.blur();
     }
-  })
+  });
 }
--- a/web/assets/copy-to-clipboard.js
+++ b/web/assets/copy-to-clipboard.js
@@ -1,16 +1,15 @@
 export default async function copyToClipboard() {
-  const copyButtons = document.querySelectorAll(
-    "button[data-copy-to-clipboard]",
-  );
-
-  copyButtons.forEach((copyButton) => {
-    copyButton.addEventListener("click", (e) => {
+  document.addEventListener("click", (e) => {
+    if (e.target.matches("button[data-copy-to-clipboard]")) {
       e.preventDefault();
+
+      const copyButton = e.target;
 
       navigator.clipboard.writeText(copyButton.dataset.copyToClipboard);
 
       const originalButtonText = copyButton.innerText;
       copyButton.classList.add("disable-click");
+      copyButton.inert = true;
       copyButton.textContent = "Copied";
 
       const screenReaderAlert = document.createElement("span");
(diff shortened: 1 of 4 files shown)


web/assets/copy-to-clipboard.js
```diff
@@ -0,0 +1,29 @@
+export default async function copyToClipboard() {
+  const copyButtons = document.querySelectorAll("button[data-copy-to-clipboard]");
+
+  copyButtons.forEach((copyButton) => {
+    copyButton.addEventListener("click", (e) => {
```
mattmachell: It's probably better to use event delegation than adding listeners to each element due to the memory footprint and not having to manage rebinding when you add/remove new elements from the DOM.

Something like

```
document.addEventListener('click', (e)=>{
  if(e.target.matches('[data-copy-to-clipboard]')){
  //your code
  }
}
```

web/assets/copy-to-clipboard.js
```diff
@@ -0,0 +1,31 @@
...
+      e.preventDefault();
+
+      navigator.clipboard.writeText(copyButton.dataset.copyToClipboard);
+
+      const originalButtonText = copyButton.innerText;
+      copyButton.classList.add("disable-click");
+      copyButton.textContent = "Copied";
+
+      const screenReaderAlert = document.createElement("span");
```
mattmachell, line 15: Nice.

web/assets/copy-to-clipboard.js
```diff
@@ -0,0 +1,31 @@
...
+  );
+
+  copyButtons.forEach((copyButton) => {
+    copyButton.addEventListener("click", (e) => {
+      e.preventDefault();
+
+      navigator.clipboard.writeText(copyButton.dataset.copyToClipboard);
+
+      const originalButtonText = copyButton.innerText;
+      copyButton.classList.add("disable-click");
```
mattmachell, line 11: You might want to add an `inert` or `disabled` attribute as well as setting the visual class.

web/assets/copy-to-clipboard.js
MikeRJustice: That's a good point. I've made that change. Thanks!

PR #992, opened 2026-05-29
Action panel success banners POC

MikeRJustice:
Adds dismissible success banners to the new action panel


web/template/warning-partial.gohtml
```diff
@@ -6,21 +6,20 @@
         {{ if eq (len .Cases) 1 }}
           {{ with index .Cases 0 }}
             <span class="govuk-!-display-none"
-              hx-get="{{ prefix (printf "/action-panel?id=%d&entity=%s" .ID (ToLower $.EntityType)) }}"
+              hx-get="{{ prefix (printf "/action-panel?id=%d&entity=%s&successMessage=%s" .ID (ToLower $.EntityType) "You have successfully created a warning.") }}"
```
Nicholassully: maybe move this "You have successfully created a warning."
Into the warning.go file like 
data.SuccessMessage = "You have successfully created a warning."
Then update the file of course.
Just thinking that way if we need more information like what was a success its easier to change and its in one place

PR #998, opened 2026-06-03
VEGA-3881: events button on action panel

almo2k:
VEGA-3881 - Add events button to action panel

What this PR does
Adds a Create Event button to the new Go action panel, following the same pattern established in VEGA-3880 (warnings).

Changes
- `server.go` — updated `/create-event` route to pass a partial template alongside the existing full-page template
- `event.go` — added `partialTmpl` parameter to `Event()`, added `DonorId`, `EntityType`, and `CaseUids` fields to `eventData`, added HTMX request detection to route between full-page and partial rendering
- `event-partial.gohtml` — new partial template that renders inside the action panel, includes all existing form fields (event type, subject, notes, file upload) with `multipart/form-data` encoding for HTMX, and reloads the action panel on success
- `action-panel.gohtml` — added Create event button using the `aw-new-event` icon
- `event_test.go` — updated all existing tests for the new three-argument `Event()` signature, updated expected `eventData` structs, added HTMX GET and POST tests
- `action-panel_cy.js` — added Cypress test for the event button

Notes
File upload is retained in the action panel partial (using `hx-encoding="multipart/form-data"`) to preserve existing user workflow. Events are always person-scoped in this context, consistent with the Angular `newEvent` behaviour.


web/template/layout/action-panel.gohtml
```diff
@@ -3,9 +3,11 @@
     {{ if eq (len .SelectedCases) 1 }}
       {{ with index .SelectedCases 0 }}
         {{ template "action-panel-button" actionPanelButton "Create warning" (prefix (printf "/create-warning?id=%d&entity=%s%s" $.DonorID (ToLower .CaseType) $.CaseUids)) "aw-create-warning" }}
+        {{ template "action-panel-button" actionPanelButton "Create event" (prefix (printf "/create-event?id=%d&entity=person%s" $.DonorID $.CaseUids)) "aw-new-event" }}
```
MikeRJustice:
```suggestion
        {{ template "action-panel-button" actionPanelButton "Create event" (prefix (printf "/create-event?id=%d&entity=%s%s" .ID (ToLower .CaseType) $.CaseUids)) "aw-new-event" }}
```
I believe that with a single case selected, we should be passing the case ID and case type into the URL here so that the event is created on the selected case and not the donor.

internal/server/event.go
```diff
@@ -65,6 +68,12 @@ func Event(client EventClient, tmpl template.Template) Handler {
 
 		data.CaseUID = ""
 		data.IsDigitalLpa = false
+		data.DonorId = entityID
```
MikeRJustice, line 71: The `entityID` will not always be the donor id. If we are creating an event for a single selected case, the `entityID` will be the case id. You could move this into the below switch statement and set the value to either the `entityID` or the `caseitem.Donor.ID` depending on the `entityType`.

almo2k: I messaged you on Slack

MikeRJustice (approved):

PR #1002, opened 2026-06-05
docker set up failing test

Nicholassully:
## Checklist


.github/workflows/build.yml
```diff
@@ -251,19 +251,34 @@ jobs:
...
+          snyk_test_args: "--severity-threshold=high"
+
+      - name: Upload Snyk Playwright scan results to GitHub Security tab
+        uses: github/codeql-action/upload-sarif@c10b8064de6f491fea524254123dbe5e09572f13 # v4.35.1
+        if: always()
+        with:
+          sarif_file: 'test-results/snyk.sarif'
+          category: snyk-docker-lpa-frontend-playwright
+
+      - name: Push Cypress Lpa-Frontend Container
```
gregtyler, line 281: I think all of this can be reverted: the Cypress container isn't pushed and the existing scan is of the main LPA Frontend container.

If you want to start pushing playwright, you'll need to add build/push steps first and create the ECR repo, but I don't think there's any need to (unlike in the E2E repo where it needs to be pulled by other services)

.github/workflows/build.yml
```diff
@@ -280,6 +295,23 @@ jobs:
...
+        env:
+          ECR_REGISTRY: ${{ steps.login-ecr.outputs.registry }}
+          END_TO_END_TESTS_ECR_REPOSITORY: sirius/lpa-frontend-playwright
+        run: |
+          docker tag 311462405659.dkr.ecr.eu-west-1.amazonaws.com/sirius/lpa-frontend-playwright:latest $ECR_REGISTRY/$END_TO_END_TESTS_ECR_REPOSITORY:${{ needs.create-tag.outputs.tag }}
+          if ${{ github.ref_name == 'main' }}
+          then
+            docker tag 311462405659.dkr.ecr.eu-west-1.amazonaws.com/sirius/lpa-frontend-playwright:latest $ECR_REGISTRY/$END_TO_END_TESTS_ECR_REPOSITORY:main-${{ needs.create-tag.outputs.tag }}
+            docker tag 311462405659.dkr.ecr.eu-west-1.amazonaws.com/sirius/lpa-frontend-playwright:latest $ECR_REGISTRY/$END_TO_END_TESTS_ECR_REPOSITORY:main-${{ needs.create-tag.outputs.tag }}-${{ needs.create-tag.outputs.timestamp }}
+
```
gregtyler, line 314: Similarly should all be removable

playwright/docker-compose.yml
gregtyler, line 1: Don't think you need this file since it's in the root docker-compose?

playwright/Dockerfile
```diff
@@ -0,0 +1,33 @@
+FROM node:25-bookworm-slim@sha256:435f3537a088a01fd208bb629a4b69c28d85deb9a60af8a710cafc3befd6e3be
+
+RUN adduser \
+    --disabled-password \
+    --gecos "" \
+    --home "/nonexistent" \
+    --shell "/sbin/nologin" \
+    --no-create-home \
+    --uid 65532 \
+    app
```
gregtyler, line 10: I think this is missing a `USER app` instruction to switch user. You'll likely run into some permissions issues when that's in, but it's worth doing properly now since we'll have to do it at some point.

playwright/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "opg-sirius-lpa-frontend",
+  "version": "1.0.0",
+  "description": "Lpa-frontend tests for the components that make up Sirius: Managed by opg-org-infra &amp; Terraform",
+  "scripts": {
+    "lint": "eslint",
+    "check-format": "prettier --check .",
+    "format": "prettier --write .",
```
gregtyler, line 8: YMMV but I've found Prettier covers the use cases of ESLint too so I would generally only use Prettier

playwright/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+WORKDIR /playwright
+
+COPY --chown=node:node package.json package.json
+COPY --chown=node:node package-lock.json package-lock.json
+
+RUN npm ci --ignore-scripts
+
+# Install chromium and dependencies
+RUN apt-get update && apt-get upgrade -y
+RUN npx --ignore-scripts -- playwright@1.62.0 install chromium --with-deps
```
gregtyler, line 21: To ensure the version of playwright is controlled, I'd suggest making this a script in package.json:
```
"install-chromium": "playwright install chromium --with-deps"
```
And calling it with `npm run`:
```suggestion
RUN npm run install-chromium
```

`npx` is generally a bit dangerous because it will install packages without warning, and the version specifier here won't be updated by dependabot/renovate

Makefile
```diff
@@ -33,6 +33,7 @@ unit-test: setup-directories
 
 build:
 	docker compose build lpa-frontend
+	cd playwright && docker compose build playwright
```
gregtyler, line 36: If you remove the extra docker-compose file:
```suggestion
docker compose build --parallel lpa-frontend playwright
```

Makefile
```diff
@@ -61,6 +62,18 @@ run-structurizr:
 	docker pull structurizr/lite
 	docker run -it --rm -p 8020:8080 -v $(PWD)/docs/architecture/dsl/local:/usr/local/structurizr structurizr/lite
 
+lint:
+	cd playwright && docker compose $(DOCKER_FEATURE_FILES) run --rm playwright lint
+
+check-format:
+	cd playwright && docker compose $(DOCKER_FEATURE_FILES) run --rm playwright check-format
```
gregtyler, line 69: Again using the top-level docker-compose file, but also DOCKER_FEATURE_FILES isn't set in this repo
```suggestion
lint:
	docker compose run --rm playwright lint

check-format:
	docker compose run --rm playwright check-format
```

PR #1011, opened 2026-06-11
Vega 3891 add new task action panel

MishNajam:
PR to add new task button to action panel

Fixes VEGA-3891

## Checklist


web/template/task-htmx.gohtml
gregtyler, line 1: This has a lot of duplication with `task.gohtml`, and it might not be obvious in the future to update both: could the form bit be shared so the templates look something like:
```html
{{ define "page" }}
  <div class="action-panel__form">
    // action panel framing stuff

    {{ template "task-form" . }}
  </div>
{{ end }}
```
and
```
{{ define "main" }}
  <div class="govuk-grid-row">
    // full view framing stuff
    {{ template "task-from . }}
  </div>
{{ end }}

gregtyler (approved):

PR #1019, opened 2026-06-17
VEGA-3964 - EPA form correct submit button on enter

MikeRJustice:
Fixes [VEGA-3964](https://opgtransform.atlassian.net/browse/VEGA-3964)

- Adds a hidden copy of the main submit button to the top of the form to ensure that the correct submit button is activated when a user presses the `enter` key while the form is focussed.
- The button copy has the tabindex="-1" and aria-hidden="true" attributes to enusure the button cannot be tabbed to and is hidden from screen readers and other accessibility tools.
- The button is hidden by setting it's dimensions and border/outline to 0 rather than using either display: none or visibility: hidden because the latter options prevent the form submitting on some browsers.

[VEGA-3964]: https://opgtransform.atlassian.net/browse/VEGA-3964?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


web/assets/main.scss
```diff
@@ -123,6 +123,15 @@ $app-action-panel-height: calc(100vh - $app-action-panel-top - 1rem);
...
 
+.app-hidden-submit {
+  width: 0;
+  height: 0;
+  margin: 0;
+  padding: 0;
+  outline: none;
+  border: 0;
+}
+
```
gregtyler: There's a GDS class called `govuk-visually-hidden` that can be used for this rather than creating a new class

gregtyler (approved):

PR #1022, opened 2026-06-17
VEGA-3952 case info

almo2k:
VEGA-3952

Implements the case info dropdown panel (info icon) on the donor header bar, following the same pattern as the existing people-info panel.

**What's done**

* `SiriusHeaderCaseInfo` handler + test (`internal/server/sirius_header_case_info.go`), mirroring `SiriusHeaderPeopleInfo`
* New `Case` fields added in `internal/sirius/case.go`: `ApplicationType`, `AttorneyDeclarationSignatureDate`, `BatchID`, `CaseRecNumber`, `LifeSustainingTreatment`, `LifeSustainingTreatmentSignatureDateA`/`B`, `LpaDonorSignatureDate`, `NoticeGivenDate`, `RegistrationDueDate`
* Filled in `sirius-header-partial-case-info.gohtml` with all fields from the ticket, conditionally rendered
* Wired the info icon button correctly in `sirius-header.gohtml` — it previously had no `id` query param and no permission gate (would 404 on click)
* Added `HasV1PersonsCasesGetPermission` to `documentPageData`, derived from the same `GetUserPermissions` call already made on page load — no extra API call
* Added `.app-case-info` SCSS block with padding ported from the original LESS (`padding: 10px 15px 15px`)
* Resolved merge conflict with main, which had added `SelectedCaseIds` to `documentPageData` — updated all affected test assertions in `document_list_test.go`
* Verified visually in browser — panel opens, fields render correctly, padding looks correct
* Cypress e2e test added at `cypress/e2e/sirius-header-case-info.cy.js`

**Deferred to QA**

* `batchId` and `caseRecNumber` field names are unverified against the real API — carried over from the Angular template. If wrong, those rows will silently not render (all fields are conditionally rendered so there's no error risk).
* LST signature date labelling needs a content decision — `Lpa.php` has `lifeSustainingTreatmentSignatureDateA` and `...DateB` as separate fields; the Angular template referenced a single unsuffixed field that doesn't exist on the entity. Both dates are rendered but B is currently labelled "Life sustaining treatment signature date B" as a placeholder. Katie to confirm correct labels.

[VEGA-3952]: https://opgtransform.atlassian.net/browse/VEGA-3952?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MikeRJustice (dismissed):

web/template/sirius-header-partial-case-info.gohtml
```diff
@@ -1,5 +1,151 @@
...
-</div>
-{{ end }}
\ No newline at end of file
+    <div class="app-case-info">
+        <dl class="govuk-summary-list govuk-summary-list--no-border">
+            <div class="govuk-summary-list__row">
+                <dt class="govuk-summary-list__key">Case owner:</dt>
+                <dd class="govuk-summary-list__value app-case-info__assignee">
+                    {{ if .Case.Assignee }}
+                        {{ .Case.Assignee.Summary }}{{ if .Case.Assignee.PhoneNumber }} ({{ .Case.Assignee.PhoneNumber }}){{ end }}
```
MishNajam: The summary isn't pulling through the team name. In the person.go file, you'll need to add something like:
`	Name                  string     'json:"displayName,omitempty"'`
and then replace .Case.Assignee.Summary with .Case.Assignee.Name

almo2k: Good spot - thanks Mish

MishNajam (approved):

PR #1031, opened 2026-06-22
VEGA-3893 allocate case to action panel

MishNajam:
PR to add allocate case to action panel. 

Additionally PR also:

- organises some of the unsorted handlers in `server.go`
- passes caseUIDs in the create task work for the action panel - as this was losing track of which case was selected after cancelling/submitting a new task without it
- adds CSS to make input search fields consistent in height with the other text fields in the action panel

Fixes VEGA-3893

## Checklist


internal/server/allocate_cases.go
```diff
@@ -95,6 +103,13 @@ func AllocateCases(client AllocateCasesClient, tmpl template.Template) Handler {
 			return err
 		}
 
+		//data.CaseUids = buildUIDQueryString(r.Form["uid[]"])
```
MikeRJustice: Left over comment.

MikeRJustice (approved):

PR #1036, opened 2026-06-24
vega-3982 adding in company ref

Nicholassully:
## Checklist


internal/sirius/person.go
```diff
@@ -11,38 +11,39 @@ type Recipient interface {
...
-	PhoneNumber           string     `json:"phoneNumber"`
-	Email                 string     `json:"email"`
-	SageId                string     `json:"sageId"`
-	CorrespondenceByPost  bool       `json:"correspondenceByPost"`
+	AlsoKnownAs           string     `json:"otherNames"`
+	Cases                 []*Case    `json:"cases,omitempty"`
+	Children              []Person   `json:"children,omitempty"`
+	CompanyName           string     `json:"companyName"`
+	CompanyReference      string     `json:"companyReference"`
+	CompanyNumber         string     `json:"companyNumber"`
```
gregtyler: I don't think you need CompanyReference any more? Also, you might consider putting this on the `Correspondent` struct rather than `Person` since it's not used by the other person types?

internal/server/create_correspondent.go
```diff
@@ -61,29 +61,30 @@ func CreateCorrespondent(client CreateCorrespondentClient, tmpl template.Templat
 		if r.Method == http.MethodPost {
 			updatedCorrespondent := sirius.Correspondent{
 				Person: sirius.Person{
-					Salutation:        postFormString(r, "salutation"),
```
gregtyler, line 64: What was the reason behind the reordering? I worry it makes it a bit harder to parse, for example since address info is no longer grouped

internal/sirius/person.go
Nicholassully: This is still needed for the time being as its being used in the add recipient in docs, which is also wrong have highlighted it with Katie and there is a ticket to also remove it from that area.
I did not want this ticket to grow in scope too much

internal/server/create_correspondent.go
Nicholassully: I think its easier to read alphabetically rather than in information groups and I know other devs in the group feel the same, but happy to change it back if you feel its better the older way

internal/sirius/person.go
gregtyler: Ah cool, I didn't notice it was there before because of the rearrange 😄 

Thoughts on moving it to the Correspondent structure?

internal/server/create_correspondent.go
gregtyler: Cool, if that's what y'all find easier 👍🏽

internal/sirius/person.go
Nicholassully: That could be a good shout and might be worth doing as part of moving the companyReference out. That way we know what just needed for the correspondent. I am thinking that the companyNumber will also be used for actor_non_case_contact which could make it a little strange to move things to the correspondent struct 🤔

gregtyler: I think it's worth making the structs as accurate as possible to avoid someone in the future trying to use `companyNumber` on a person type that doesn't support it

Nicholassully: Ok cool I will move it to that struct 👍

gregtyler (dismissed):

gregtyler (approved):

PR #1047, opened 2026-07-02
VEGA-3894 assign task

almo2k:
VEGA-3894

Added Assign Task button to the Go action panel, following the same pattern established in VEGA-3883. 
The handler and full-page template already existed; this makes it HTMX-friendly and wires it into the action panel.

**Changes**

`action_panel.go` — adds `TasksForCase` to `ActionPanelClient`, fetches the selected case's outstanding tasks alongside the existing draft count call, and adds Assign Task to `GetActionPanelButtons()`. Disabled unless exactly one case is selected and it has at least one outstanding task. URL passes a repeated `id` param per task, plus `donorId` and `uid[]`.

`document_list.go`, `view_document.go` — both call `GetActionPanelButtons()` directly (outside `ActionPanel`), so both needed the same `TasksForCase` addition to keep the signature change compiling.

`assign_task.go` — adds `partialTmpl` parameter, `DonorID`, `CaseUids`, `CaseType` and `TaskIDs` fields to `assignTaskData`, and HTMX request detection to route between full-page and partial render.

`assign-task-wrapper.gohtml`, `assign-task-partial-wrapper.gohtml` — form fields inlined directly in both, matching the `change-status` pattern rather than a shared block.

`server.go` — updates the `/assign-task` route to pass both templates.

`action_panel_test.go`, `document_list_test.go`, `view_document_test.go`, `assign_task_test.go` — updated expected `ActionPanelButtons` slices and `TasksForCase` mocks; new HTMX tests added for GET and POST via partial template.

`action-panel.cy.js` — adds Cypress test for the Assign Task button.

**Behaviour**

- No cases selected → button disabled
- One case selected, no outstanding tasks → button disabled
- One case selected, with outstanding tasks → button enabled, opens Assign Task form in action panel
- Multiple cases selected → button disabled


web/template/assign-task-partial-wrapper.gohtml
```diff
@@ -0,0 +1,81 @@
...
+
+    {{ template "error-summary" .Error }}
+
+    {{ if .Success }}
+      <span class="govuk-!-display-none"
+            hx-get="{{ prefix (printf "/action-panel?id=%d&entity=%s%s" .DonorID .CaseType .CaseUids) }}"
+            hx-target=".action-panel__content"
+            hx-swap="innerHTML"
+            hx-trigger="load"
+      ></span>
```
davidbridgwood: see Mikes recent PR: https://github.com/ministryofjustice/opg-sirius-lpa-frontend/pull/1053/changes we're triggering a refresh on a successful form submission now.

cypress/e2e/action-panel.cy.js
davidbridgwood, line 1: the "displays the edit epa button on the action panel" needs a mock for the tasks endpoint too since it uses a different case than the rest of the tests (it needed to be an EPA to be selectable)

web/template/assign-task-partial-wrapper.gohtml
```diff
@@ -0,0 +1,81 @@
...
+              <label class="govuk-label govuk-radios__label" for="f-assignTo-2">
+                Team
+              </label>
+            </div>
+            <div class="govuk-radios__conditional govuk-radios__conditional--hidden" id="conditional-assignTo-2">
+              {{ template "select" (select "assigneeTeam" "Team" nil .Error.Field.assigneeTeam (options .Teams)) }}
+            </div>
+          </div>
+        </fieldset>
+      </div>
```
davidbridgwood, line 63: I wonder if there is enough duplication here to pull out into a separate template as we've done in other cases. Don't feel massively strongly about it though so happy to stick with it as is if you disagree.

almo2k: I'm thinking we stay with it as is for now but definitely a good case for a future refactor

davidbridgwood (approved):

PR #1054, opened 2026-07-07
Vega 3994 add investigation button action panel

MishNajam:
PR adds investigation button to the action panel. Button is only enabled when one case is selected. Also:

- hides the investigation button depending on permissions as per [current angular code](https://github.com/ministryofjustice/opg-sirius/blob/e4fd2fd9ef72224acaa7cac9ab868a0452fa8ac6/front-end/lpa/src/app/js/action-widget/opgActionWidget.module.js#L93)
- Adds the investigation icon 
- Removes unnecessary query param in url in create relationship form (previous code review feedback)

Fixes VEGA-3994 #patch

## Checklist


web/template/create-investigation-wrapper.gohtml
davidbridgwood, line 1: i think its worth renaming this file to `create-investigation-wrapper.go` to follow the convention we've established

davidbridgwood (approved):

PR #1062, opened 2026-07-13
VEGA-4003 : make doc-lists in compare view scrollable

davidbridgwood:
[vega-4003](https://opgtransform.atlassian.net/browse/VEGA-4003)

### DIFF
--- a/web/assets/main.scss
+++ b/web/assets/main.scss
@@ -547,6 +547,11 @@ $app-action-panel-height: calc(100vh - $app-action-panel-top - 1rem);
   flex: 1;
 }
 
+.document-list-scrollable {
+  max-height: 100vh;
+  overflow-y: auto;
+}
+
 .govuk-width-container:has(> .action-panel) {
   display: grid;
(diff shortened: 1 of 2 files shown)


MikeRJustice (dismissed):

web/assets/main.scss
```diff
@@ -1145,3 +1150,7 @@ $app-action-panel-height: calc(100vh - $app-action-panel-top - 1rem);
     border-bottom: 1px solid govuk-functional-colour(border);
   }
 }
+
+.govuk-table[data-module="moj-sortable-table"] + div[role="status"] {
```
MikeRJustice, line 1154: There was an absolutely positioned element added to the dom by the moj sortable table component that was pushing the height of the table container down. I've given it a bottom value of 0 so it sticks to the bottom of the docs table and therefore removes the extra space.  The element is just being used as an accessibility label for the sort status of the table, so it is visually hidden.

MikeRJustice (approved):
Approving my changes on David's behalf

PR #1066, opened 2026-07-15
VEGA-4011: large font action panel

almo2k:
VEGA-4011

Implements the large-font accessible theme for the new HTMX action panel, matching the existing behaviour on the legacy Angular action widget.

The old Angular app swaps the entire #themeCSS stylesheet bundle when an accessible theme is selected. The new Go/HTMX frontend is a separate top-level page (not an iframe), with a single compiled stylesheet, so that mechanism doesn't apply directly. 
Instead, this follows the existing pattern already used for dark mode: load-classes.js reads the siriusTheme cookie on page load (before paint) and adds a class to <html>, which SCSS then targets.

Note
This covers the action panel only, per the ticket. It doesn't implement the site-wide large-font styling (base font-size bump, timeline, search dropdown, etc.) that the old Angular app applies via accessible-base.less

### DIFF
--- a/web/assets/load-classes.js
+++ b/web/assets/load-classes.js
@@ -15,3 +15,7 @@ if (
 ) {
   document.documentElement.className += " app-!-html-class--dark";
 }
+
+if (document.cookie.indexOf("siriusTheme=accessible-") > -1) {
+  document.documentElement.className += " app-!-html-class--large-font";
+}
--- a/cypress/e2e/action-panel.cy.js
+++ b/cypress/e2e/action-panel.cy.js
@@ -619,4 +619,15 @@ describe("Action Panel", () => {
     cy.get(".action-panel__form").should("exist");
     cy.get(".action-panel__form").contains("Assign Task");
   });
+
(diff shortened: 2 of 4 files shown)


web/assets/load-classes.js
```diff
@@ -15,3 +15,7 @@ if (
 ) {
   document.documentElement.className += " app-!-html-class--dark";
 }
+
+if (document.cookie.indexOf("siriusTheme=accessible-") > -1) {
```
MikeRJustice, line 19: It might be worth including `accessible-dark` in this check considering we are currently also supporting dark mode?

almo2k: Hi mate, this should cover all accessible modes — confirmed in local testing. The check is a substring match on accessible-, so accessible-dark (and every other colour variant) already satisfies it without needing a separate case

MikeRJustice: Ah I see. All good then!

MikeRJustice (approved):

PR #1072, opened 2026-07-24
VEGA-4033 - Add hover text to header buttons

MikeRJustice:
Fixes [VEGA-4033](https://opgtransform.atlassian.net/browse/VEGA-4033)

[VEGA-4033]: https://opgtransform.atlassian.net/browse/VEGA-4033?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cypress/e2e/compare-document.cy.js
+++ b/cypress/e2e/compare-document.cy.js
@@ -333,4 +333,28 @@ describe("Compare documents", () => {
           });
       });
   });
+
+  it("should display the header button labels on hover", () => {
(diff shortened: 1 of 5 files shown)


MishNajam:
I think the label needs to be updated for the back to timeline button (Needs to be Return)

Otherwise, wonder if it would be handy to add a cypress test for these

MishNajam (approved):

PR #1076, opened 2026-07-27
VEGA-4021: Create a LPA

almo2k:
## VEGA-4021

Migrates the Create LPA case flow from Angular to Go/HTMX, following the pattern established by the earlier EPA migration.

### Files changed

- `internal/sirius/lpa.go` **(new)** — `Lpa` struct (embeds `Case`, adds LPA-only fields) plus `Client.Lpa`/`CreateLpa`/`UpdateLpa`, mirroring `epa.go`'s shape exactly.
- `internal/sirius/lpa_test.go` **(new)** — Pact consumer contract tests for the three new client methods, mirroring `epa_test.go`.
- `internal/server/create_lpa.go` — full GET/POST handler.
- `internal/server/create_lpa_test.go` — 24 tests.
- `web/template/create-lpa.gohtml` — full form markup across all four accordion sections.
- `web/template/create-lpa-wrapper.gohtml`, `create-lpa-partial-wrapper.gohtml` — two pre-existing bugs fixed
- `internal/server/server.go` — one-line addition, `CreateLpaClient` was missing from the composite `Client` interface.

### What's built

**Section 1 (2–4): Attorney and appointment type** — LPA type, application type (+ A-reference reveal), receipt date (permission-gated on `v1-lpas-edit-dates`/PUT), appointment type (4 options), Attorneys/Replacement attorneys cards.

**Section 2 (5–8): LPA instructions and notified people** — conditional card that swaps between "When can the LPA be used?" (PFA) and "Life sustaining treatment (LST)" (HW) based on LPA type; Notified person; Preferences and instructions (3 checkboxes, "None" clears the other two, per Mikes GDS "add option for none" comment on the ticket).

**Section 3 (9–11): Signatures and Certificate Provider details** — one combined card per the Figma (originally scaffolded as two separate cards — merged to match): donor + signature date, certificate provider(s) + per-provider date, attorneys/replacement attorneys + per-attorney date.

**Section 4 (12–15): Application form** — Applicant (donor/attorney + per-attorney checkboxes + date signed), Correspondent, Application fee (4 checkboxes: card, cheque, reduced fee, repeat application — "reduced fee" reveals an Exemption/Remission radio), Any other information.

**Edit path** — `isEditing`/`UpdateLpa` support is built in alongside create, even though KWE ticket comment scopes editing an existing LPA as a *separate* ticket. I add it in deliberately as it gives the edit ticket a head start.

### Deliberately deferred (blocked on actor creation)

`/create-attorney` is currently hard-coded to EPA cases only (`CreateAttorneyClient.Epa(...)`), and there's no equivalent route for notified persons/certificate providers/correspondents for LPA at all. So every "Add X" button (attorney, replacement attorney, notified person, certificate provider, correspondent) is rendered but disabled — visible per the ticket's own "could add the button, but adding actors will be implemented separately" wording, not functional. Existing records (on an edit) are still displayed read-only where the API already returns them.

### Structural decisions

- **LPA-only fields live on `sirius.Lpa`, not `sirius.Case`.** Initially added several LPA-specific fields directly onto the shared `Case` struct; corrected once `epa.go` was seen in full — EPA doesn't cram its own fields into `Case` either, it embeds `Case` and adds its own on top. Fields genuinely shared across case types (`ReceiptDate`, `CaseAttorneySingular` etc.) were already on `Case` before this ticket and are untouched.
- **Certificate provider / attorney signature dates are not on `sirius.Lpa`.** The Figma shows each CP/attorney with its own date, which a single case-level field can't represent. Rendered as disabled placeholders in the template; the real field needs to live on the actor record itself, as part of the actor-creation ticket. `digital_lpa.go`'s `LpaStoreAttorney.SignedAt` / `LpaStoreCertificateProvider.SignedAt` confirm per-record dates are an established pattern elsewhere in the codebase, for what it's worth as precedent (that's a different API — the digital LPA store — so it's precedent for the *concept*, not a direct fix).
- **`LifeSustainingTreatmentSignedAndWitnessed` is an invented field/JSON key.** The Figma HW form has a "Signed and witnessed correctly" Yes/No that has no corresponding property anywhere in `Lpa.php` or the classic-case entity. Kept as a `*bool` (closest literal match to the Figma checkbox). Needs confirming against the real API — may need an API change to support at all.
- **`repeatApplicationReference` is shown in the UI but cannot currently be saved.** The PHP `CreateLpa` command class has no property/getter for it (only the entity does) — confirmed by its absence from `Lpa.php`, and consistent with the legacy Angular form marking the equivalent field `readonly: true`. The Figma shows it as a live, editable field, which is a genuine contradiction between design and current API capability, not something resolvable from the frontend side.
- **Certificate provider count limit (1 vs 2) defaulted to the permissive value (2).** The design annotation says "older LPAs can have 2, newer LPAs can only add 1" — this form has no way to know an LPA's age at creation time, so it currently always allows up to 2. Worth a decision on whether/how to determine "age" here.
- **`UpdateLpa`'s endpoint (`/lpa-api/v1/lpas/{id}`, PUT)** is inferred by direct analogy with EPA's `/lpa-api/v1/epas/{id}` and is consistent with `permissions_global.php`'s `private-case-update` group, but hasn't been confirmed against an actual `UpdateLpa.php` command file.


web/template/layout/create-lpa.gohtml
```diff
@@ -63,10 +146,27 @@
         <div class="govuk-summary-card">
           <div class="govuk-summary-card__title-wrapper">
             <h2 class="govuk-summary-card__title">
-              When can the LPA be used?
+              {{ if eq .Lpa.SubType "hw" }}Life sustaining treatment (LST){{ else }}When can the LPA be used?{{ end }}
             </h2>
           </div>
           <div class="govuk-summary-card__content">
+            {{ if eq .Lpa.SubType "hw" }}
```
MikeRJustice, line 189: At the point of creating a new LPA, this check will always return false even if the user has ticked the HW box resulting in the PFA section always being displayed.

MikeRJustice (dismissed):

MikeRJustice (approved):

PR #1088, opened 2026-07-30
VEGA-4022 - Implement create attorney on LPA form

MikeRJustice:
Fixes [VEGA-4022](https://opgtransform.atlassian.net/browse/VEGA-4022)

[VEGA-4022]: https://opgtransform.atlassian.net/browse/VEGA-4022?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cypress/e2e/create-attorney.cy.js
+++ b/cypress/e2e/create-attorney.cy.js
@@ -1,10 +1,9 @@
-const fillInAttorneyDetails = () => {
+const fillInAttorneyDetails = (caseType = "epa") => {
   cy.get("#f-salutation").type("Prof");
   cy.get("#f-firstname").type("Melanie");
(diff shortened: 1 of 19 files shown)


almo2k (dismissed):
Approved, I noticed the change from htmx:afterSwap to htmx:afterSettle, Just flagging it for visibility — this means the MOJ/GOV.UK reinit and action-panel width toggle now run once the DOM has settled (adds htmx's default ~20ms delay) rather than immediately on swap. I checked the CSS and this doesn't interact badly with the action-panel's transitions, so looks fine to me but its something to be mindful of it we start to play with transitions.

almo2k (approved):

PR #1097, opened 2026-08-04
vega-4088 change search from 15 to 20 #patch

Nicholassully:
## Checklist


web/template/search.gohtml
```diff
@@ -34,9 +37,6 @@
                     </div>
                 </div>
             {{ else }}
-
-                <p class="moj-pagination__results govuk-!-padding-0 govuk-!-margin-bottom-6">Showing <b>{{ .Pagination.Start }}</b> to <b>{{ .Pagination.End }}</b> of <b>{{ .Total }}</b> results for <b>{{ .SearchTerm }}</b></p>
```
gregtyler, line 38: From the screenshot on the ticket, it looks like this should still be here, it's just the header that should change?

Nicholassully: From the screenshot (the second one) I can see that the `Showing x to y of z results for ABC`. Has been removed in less I am looking at the wrong thing

gregtyler: Ah sorry, I'd missed that this was in two places and you'd removed the one above the table. I guess the Cypress test steps don't need to be removed since that text should still be below the table, but I'm not sure they were providing much value 👍🏽

gregtyler (approved):

PR #1100, opened 2026-08-04
VEGA-4074 -  Update LPA form sections on case subtype selection

MikeRJustice:
Fixes [VEGA-4074](https://opgtransform.atlassian.net/browse/VEGA-4074)

- Add Javascript module to toggle the LPA form sections depending on the users choice of case subtype (HW/PFA)
- Add unit test

[VEGA-4074]: https://opgtransform.atlassian.net/browse/VEGA-4074?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web/assets/lpa-form-subtype.js
+++ b/web/assets/lpa-form-subtype.js
@@ -27,8 +27,6 @@ export default function lpaFormSubtype(scope) {
 
     hwSection.hidden = value !== "hw";
     pfaSection.hidden = value !== "pfa";
-
-    console.log(pfaSection.hidden);
   }
 
   handleFormSubtype();
--- a/web/assets/lpa-form-subtype.js
+++ b/web/assets/lpa-form-subtype.js
@@ -1,0 +1,33 @@
+export default function lpaFormSubtype(scope) {
+  scope = scope || document;
+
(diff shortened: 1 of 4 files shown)


almo2k (dismissed):
Nice and clean overall. 

Unless I'm missing something I think there's a gap in the test — all three tests check the radio, then dispatch a change event, But None of them test what happens when a radio is already checked before lpaFormSubtype() runs (i.e. edit mode, where the subtype's pre-selected on page load and no change event fires). 

If you agree with my understanding then it might be worth adding a test for that case to cover it all off?

web/assets/lpa-form-subtype.js
```diff
@@ -27,6 +27,8 @@ export default function lpaFormSubtype(scope) {
 
     hwSection.hidden = value !== "hw";
     pfaSection.hidden = value !== "pfa";
+
+    console.log(pfaSection.hidden);
```
almo2k: Is this a leftover debug?

almo2k (dismissed):

almo2k (approved):

PR #1104, opened 2026-08-06
Vega 4026 lpa add notified people

davidbridgwood:
[vega-4026](https://opgtransform.atlassian.net/browse/VEGA-4026)

- create and edit notified-persons
- implements with the new single file for templates pattern
- only allow max of 4 notified persons to be created.

### DIFF
--- a/internal/server/create_notified_person.go
+++ b/internal/server/create_notified_person.go
@@ -70,7 +70,8 @@ func CreateNotifiedPerson(client CreateNotifiedPersonClient, tmpl template.Templ
 			return err
 		}
 
-		data.AllowNewNotifiedPerson = allowNewNotifiedPerson(len(lpa.NotifiedPersons))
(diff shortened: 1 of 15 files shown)


internal/server/create_notified_person.go
```diff
@@ -0,0 +1,170 @@
...
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/ministryofjustice/opg-go-common/template"
+	"github.com/ministryofjustice/opg-sirius-lpa-frontend/internal/sirius"
+)
+
+// TODO: fix this cause its wrong when you're adding the final notified person
```
MikeRJustice: Does this still need to be done?

davidbridgwood: oops, nope did fix that will remove.

davidbridgwood: and turns out from Katie the max is actually meant to be 5

